Issue #4784, opened 2016-03-10
Remove old poppler/ghostscript references


michaelrsweet:
Version: -feature
CUPS.org User: computersforpeace

There are still a few stray macros defined in cups for tools that are not used in the main cups source (poppler and ghostscript). Might as well remove them?

Patch attached.

michaelrsweet:
[master f73476a] Drop unused definitions.

PR #4787, opened 2016-03-17
--enable-debug-printfs bugfix + other cosmetic fixes

computersforpeace:
Addresses these:

https://github.com/apple/cups/issues/4784
http://www.cups.org/pipermail/cups/2016-March/027577.html

I hope you're taking Github PRs now?

Regards,
Brian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.o
 *.so
 *.so.*
+*.swp
 .buildrev
 autom4te.cache
 config.h
--- a/config.h.in
+++ b/config.h.in
@@ -512,24 +512,6 @@
 
 
 /*
- * Location of the poppler/Xpdf pdftops program...
- */
-
-#undef HAVE_PDFTOPS
-#undef HAVE_PDFTOPS_WITH_ORIGPAGESIZES
-#define CUPS_PDFTOPS	"/usr/bin/pdftops"
-
-
-/*
- * Location of the Ghostscript gs program...
- */
-
-#undef HAVE_GHOSTSCRIPT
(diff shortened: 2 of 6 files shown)


michaelrsweet:
Yes, we'll be accepting PRs, but please be patient as I am currently swamped making this Github thing happen... :)

michaelrsweet:
The poppler/ghostscript changes were addressed separately.

Fixed debug_ops.

[master 508de67] The debug strings table for mime type operations was out of sync with the enumeration (PR #4787)
[branch-2.1 f94124c] Mirror fix from master.

I'm undecided on whether I'll add the vim swap file change to .gitignore; that's something you can handle locally in the meantime...

computersforpeace:
Did you have any comments on this part of my original question?

> It'd be best if we could use the compiler to our advantage to catch
these types of errors (e.g., introduce some kind of compile-time
assertion;, or use a switch/case without a 'default' instead of an
array, since many compilers would warn about the unused enum), but I
didn't really want to refactor this all for a bugfix. I'm also not as
familiar with the non-Linux systems that CUPS is built for, so I'm not
sure what (if any) C99 features I could use, for instance.

I notice you didn't even take my comment for the enum:

    /* NOTE: update debug_ops[] in type.c whenever editing this enum */

so it seems this is just as liable to get broken again in the future.

michaelrsweet:
mime.h is a public header so I don't want to expose internal implementation details there (probably just confuse anyone reading the header).

But regardless, we've added 2 new operations to the enum over the last 18 years so I'm not particularly worried about it.

PR #4807, opened 2016-04-19
Fix spanish gerunds in index.html

OdyX:
See the Debian bug: https://bugs.debian.org/821788 from Santiago Vila:

> The file doc/es/index.html.in is a translation of doc/index.html.in, but gerunds like "Using Network Printers" have been translated as "Usando [...]" (Spanish gerund, which is not the same as English gerund). English gerund is a noun, so it should be translated as such.


michaelrsweet:
Thanks!

PR #4808, opened 2016-04-20
fix preprocessor if for dnssd || avahi

justicz:
took entirely too long to figure out why this wasn't compiling :)


michaelrsweet:
[master 5558a2e] PR #4808: Fix builds without Avahi or mDNSResponder.

PR #4811, opened 2016-04-21
get_printers: check SystemGroup members

### LINKED ISSUE #4810: restricting access to printer by user hides printer from list
step to reproduce (tested on CUPS 2.1.3):
1. go to http://localhost:631/printers/myprinter (login as SystemGroup/"lpadmin" member)
2. select "Administration" - "Set Allowed Users"
3. fill "Users:" field with some usernames: say "zzz" (don't include youself or "lpadmin" group)
4. select "Allow these users to print" and press "Set Allowed Users"
check /etc/cups/printers.conf:
...
AllowUser zzz
...
run "lpstat -a" - does not show "myprinter".
go to  http://localhost:631/printers/ - does not show "myprinter".
but "myprinter" directly accessible through http://localhost:631/printers/myprinter

dab18:
resolve #4810
Signed-off-by: Dmitry Bakshaev <dab1818@gmail.com>

### DIFF
--- a/scheduler/ipp.c
+++ b/scheduler/ipp.c
@@ -11089,6 +11089,11 @@ user_allowed(cupsd_printer_t *p,	/* I - Printer or class */
   pw = getpwnam(username);
   endpwent();
 
+  int i;
+  for (i = 0; i < NumSystemGroups; i ++)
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Behaves correctly.

PR #4829, opened 2016-05-29
Fix typo in cupsctl man webpage

leoarnold:
The typo / conversion error seems to have snuck in with CUPS 2.0


michaelrsweet:
The change needs to actually happen in the man page source and not the formatted HTML from it. Investigating...

michaelrsweet:
[master 3bd2e63] Fix typo in cupsctl man page (PR #4829)
[branch-2.1 91480df] Mirror fix from master.

Issue #4831, opened 2016-06-02
Unable to build .deb


mfly:
packaging/cups.list.in is not up-to-date which results in ```$ make deb ``` failing.
Build log: https://gist.github.com/mfly/fc6f1487afefbf77a833326ecb6f0c34.

It looks like this issue: https://github.com/apple/cups/issues/4518 was solved only partially and banner files still exist in cups.list.in. There is a similar problem with files in doc/help/.

michaelrsweet:
Closing this one, tracking with PR #4832.

PR #4832, opened 2016-06-02
Fix for EPM packaging file

mfly:
https://github.com/apple/cups/issues/4831


michaelrsweet:
OK, the documentation files change isn't correct - we separate API help into the -devel package.

michaelrsweet:
Also found a bunch of missing files.

[master 63c5db7] Fix EPM packaging file.

Issue #4835, opened 2016-06-08
Malformed device-id for dnssd discovered printers


bigon:
Hi,

I'm running cups 2.1.3 (on debian unstable if it matters)

When discovering printers, I think that the device-id returned by the server is malformed for the ones discovered by dnssd:
 
```
device-uri: dnssd://HP%20LaserJet%20P2055dn%20%5B447A47%5D._pdl-datastream._tcp.local/
device-id: MFG:HP;MDL:LaserJet P2055dnCMD:PS,PCL;
```

For the same printer I get also:

```
device-uri: socket://192.168.1.137:9100
device-id: MFG:Hewlett-Packard;MDL:HP LaserJet P2055dn;CMD:PJL,POSTSCRIPT,PCL,PCLXL;CLS:PRINTER;DES:HP LaserJet P2055dn;FWVER:20100308;
```

As you can see, there is a semicolon missing in the first case after the name of the printer

PR #4836, opened 2016-06-08
Fix malformed device-id for printers found by dnssd

bigon:
Add missing semicolon between printer model and command

Closes: #4835


michaelrsweet:
Also committing to 2.1.x.

PR #4839, opened 2016-06-15
Editing cups-lpd.man.in

zdohnal:
Hello Mike,

Jirka (jpopelka) changed cups-lpd.man.in a little in our project, but his patch is for old version of cups (1.6.3), so I tried to edit his patch into format, which can be applied into current version of CUPS. Do you want to merge it into your project?


michaelrsweet:
While that might be appropriate advice on a specific Linux distro, CUPS is still used on platforms that use inetd and xinetd so I am not comfortable removing that information.

PR #4858, opened 2016-07-22
Improved German translation

robert-scheck:
Improved German translation: Added missing translations, corrected wrong wordings, corrected typos, used better phrases, corrected English-German mixtures, etc.

In case you disagree partially with my suggestions, please at least cherry-pick everything else and come back to me to discuss the leftovers.

Personally, I do not treat this patch as a "significant contribution to CUPS" - if I'm wrong, please let me know (the patch adds no new features, no large codebase changes, just a bunch of translation updates).


cwickert:
As native German speaker, someone who studied linguistics and has managed various German translation teams (Xfce, LXDE, …) , I strongly suggest accepting this patch.

michaelrsweet:
@robert-scheck : Translations are not treated as significant contributions, so we are good there.

@cwickert : No need to push hard for this, as long as a translation is complete we take it... :)

robert-scheck:
The translation might not be 100% complete, but should be > 95% from my point of view. There are some things where I'm not sure how to translate (e.g. `command.tmpl`) and maybe some further things that slipped through. In case you see some issues with the patch, let me know early so that I can pay attention to it. Btw, I don't think @cwickert wanted to push hard in an unfriendly way, but more make clear that the overall quality of the translation gets improved due to the patch (at least that's my understanding).

Issue #4860, opened 2016-07-27
Extend buffer size for name of loaded PPDs up to 256 characters


mariospr:
We've recently received the [report of an user](https://groups.google.com/d/msgid/endless-community/020a9ed9-55d9-44ed-a6aa-508396c35c5b%40googlegroups.com) of [Endless](https://endlessm.com) mentioning that he could not install his Epson Stylus Photo T60 printer and, after some investigation, I found the 128 characters limit was the reason for the failure, so I wonder if that could be raised and such a patch accepted upstream, which would be ideal.

For extra context, we have a system in place that automatically downloads and installs drivers from Openprinting.org, making them discoverable to CUPS via paths under `/usr/share/ppd/eos-config-printer`which, in the case of the printer reported by our user would look something really long: 

`/usr/share/ppd/eos-config-printer/epson-inkjet-printer-stylus-photo-t50-series_ppds_Epson/Epson-Stylus_Photo_T60_Series-epson-driver.ppd.gz`. 

This, when translated into a virtual path, would look something like this:
`lsb/usr/eos-config-printer/epson-inkjet-printer-stylus-photo-t50-series_ppds_Epson/Epson-Stylus_Photo_T60_Series-epson-driver.ppd.gz`

...which would be 132 characters long and therefore would get truncated, making CUPS fail with a `Unable to open "/usr/[...]T60_Series-epson-driver.p" - No such file or directory` error.

I understand that the fact we have such a looong pathname to point to there the PPD files will end up living is unfortunate and probably suggests that a patch like that should live downstream only, but before assuming so or looking into other options I thought I would ask, as perhaps it's not a terrible idea either to raise this buffer size a bit (e.g. to 256 characters, to keep powers of 2 sizes) upstream?

mariospr:
Just send a Pull Request for this: https://github.com/mariospr/cups/commit/759c9b49003bb2a07f2b4db38a1ae4ac58d2617b

Please let me know what you think, and feel free to merge it right away if it's good enough. Thanks!

michaelrsweet:
The maximum allowed size of a "name" value in IPP is 255 octets (bytes), so that's as far as we can go... I'll need to review things to make sure it doesn't cause any regressions.

mariospr:
Thanks @michaelrsweet, for looking into this so quickly. 255 bytes would be certainly more than enough even for our particular case so that should be fine. Please let me know what you think once you review it, and thanks again

PR #4861, opened 2016-07-27
Increase the allocated buffer size for names of loaded PPDs (Issue #4860)

mariospr:
Having 128 characters is probably enough in most systems, but in Endless
we have a custom system in place to automatically download printer drivers
that will be accessible from long paths under /usr/share/ppd/eos-config-printer
(e.g. /usr/share/ppd/eos-config-printer/epson-inkjet-printer-201301w_ppds_Epson)

As this path will become part of the virtual path for the PPD, this limit
will cause trouble in cases where the full PPD name exceeds it, since the
path CUPS will try to open will not exist, as it will be truncated.


michaelrsweet:
[master 4a971b9] Update cups-driverd and spec-ipp to be consistent about supporting name(255) values for the "ppd-name" attribute (PR #4860)

PR #4879, opened 2016-09-19
Return NULL if no connection available. Fix issue #4870

xuzhen:


michaelrsweet:
This "fix" will break connections to network printers/servers with more than one address, which is basically all of them (IPv4+IPv6). But I'll keep investigating...

PR #4895, opened 2016-10-06
Update cups.desktop.in

EvertonMelo:
adding interface for Portuguese Portugal.
adding interface for Korean
adding interface for Thai
adding interface for Arabic
adding interface for Uzbek

support on:
https://support.mozilla.org/en-US/kb/localization-dashboard
standard ISO 639:
https://en.wikipedia.org/wiki/ISO_639


EvertonMelo:
Thank you for accept my Pull Request

Issue #4872, opened 2016-08-23
Fix cross-compilation issues


Redfoxmoon3:
Currently when cross compiling, it fails on trying to run 'genstrings' which gets cross-compiled, that's obviously incorrect... edit: same with mantohtml.

michaelrsweet:
CUPS currently does not support cross-compilation.

michaelrsweet:
I believe this issue has been addressed through the changes for Issue #4897

PR #4897, opened 2016-10-09
Cross-build fixes

OdyX:
Helmut Grohne has reported in https://bugs.debian.org/837936 the only two missing pieces to allow CUPS cross-building. Here come the patches.

This addresses #4872.

### DIFF
--- a/Makedefs.in
+++ b/Makedefs.in
@@ -18,6 +18,7 @@
 AR		=	@AR@
 AWK		=	@AWK@
 CC		=	@LIBTOOL@ @CC@
+CC_FOR_BUILD	=	@CC_FOR_BUILD@
 CHMOD		=	@CHMOD@
 CXX		=	@LIBTOOL@ @CXX@
 DSO		=	@DSO@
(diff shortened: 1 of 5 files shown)


michaelrsweet:
We can't accept these changes as-is.

1. Compiler tests should be in config-scripts/cups-compiler.m4
2. I'm not crazy about the mantohtml changes since strncpy usage will throw a LOT of warnings on many platforms and with most code checking tools.
3. Completely dropping genstrings here means that changes to the driver info file will never be reflected in the message catalogs.

I'd be more interested in changes that bypass building/running mantohtml and genstrings when doing cross-compilation but not when doing a normal build. The HTML versions of the man pages are part of the repository (don't need to be rewritten normally), and similarly we have current message catalogs in the repository.

michaelrsweet:
OK, I've put together my own changes that should fix things for cross-compilation.  Please test and let me know if you run into any problems...

[master 105922ec6] Add support for cross-compilation (Issue #4897)

Redfoxmoon3:
config-scripts/cups-pam.m4. line 56	"marwin*)" typo?

michaelrsweet:
Fixed!

PR #4903, opened 2016-10-17
Localization for job's and printer's status messages and messages in cupsdAddEvent()

dimv36:
Those patch add localization for messages in cupsdAddEvent() and localization for status of printers and jobs.

### DIFF
--- a/scheduler/ipp.c
+++ b/scheduler/ipp.c
@@ -181,7 +181,7 @@ cupsdProcessIPPRequest(
     */
 
     cupsdAddEvent(CUPSD_EVENT_SERVER_AUDIT, NULL, NULL,
-                  "%04X %s Bad request version number %d.%d",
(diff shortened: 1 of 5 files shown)


dimv36:
Sorry, those patch add only translation possibility of those messages, but not translation at all.

dimv36:
This commit adds localization support in functions cupsdAddEvent() & cupsdSetJobState()

michaelrsweet:
OK, so as-is this patch is not usable.

One of the reasons we have not localized the notification text in cupsd is that the subscriber's natural language can be different than the system-wide default language. Doing this properly requires localizing the event multiple times (once for each subscription).

michaelrsweet:
Also, the cupsNotifyGetSubject and cupsNotifyGetText functions in libcups localize notifications on the client-side (independent of the subscription language)

michaelrsweet:
Sorry, not accepting this PR...

Issue #4906, opened 2016-10-18
CUPS printers question


dimv36:
Hello,

Printers can have 2 type: shared and local. If we set CUPS to work on TCP (e.g. set ServerName in client.conf), then CUPS will give only shared printers on CUPS-Get-Printers request.

Next, in those configuration I can try to create printer with name, that already exists in CUPS (local printer, that is not shown by CUPS-Get-Printers request), CUPS execute add_printer() function. 

CUPS checks, that printer exists and updates configuration of local printer. Is it correct?

michaelrsweet:
That is an unfortunate side-effect of how printer sharing works, and as long as you have remote administration enabled this is something you can do (configure a printer but not have access to it remotely until you mark it as shared...)

PR #4907, opened 2016-10-19
Fix side effects of remote administration on local printers

dimv36:
This is fix of remote administration's side effects of updating local printer's configuration, that we discussed in issue (https://github.com/apple/cups/issues/4906). I suppose, if CUPS running in not-local mode and printer is local we deny updating printer's configuration.

### DIFF
--- a/scheduler/ipp.c
+++ b/scheduler/ipp.c
@@ -2237,6 +2237,7 @@ add_printer(cupsd_client_t  *con,	/* I - Client connection */
 		need_restart_job,	/* Need to restart job? */
 		set_device_uri,		/* Did we set the device URI? */
 		set_port_monitor;	/* Did we set the port monitor? */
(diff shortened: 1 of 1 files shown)


michaelrsweet:
*But* if you have enabled remote administration then you will still want to be able to modify local (not shared) print queues remotely.

dimv36:
I suppose, if I can't get local printers via CUPS-Get-Printers request with enabled remote administration, then I can't modify local print queues remotely. 

Do you want modify local print queue? OK, disable remote administration.

Or another solution. We can't get any local print queues with enabled remote administration. So, may we remove any checks on printer (does it local) and we will got full list of printers (include local). Then your idea about

> But if you have enabled remote administration then you will still want to be able to modify local (not shared) print queues remotely.

seems to me more correct then we have.

PR #4925, opened 2016-11-22
Make ZPL Orientation configurable 

frazhome:
shouldn't the orientation configurable? Sometimes I need the label 'leading edge' and sometime 'foremost edge'


frazhome:
sorry for some garbage. It is my first  pull request.

Only the commit with ZPL Orientation configurable ... is usefull

michaelrsweet:
The issue is that the leading edge of the output is *supposed* to be the top of the page image when orientation-requested=3 (portrait) - that's the IPP and Printer MIB definition. To get the old behavior you need to use orientation-requested=6 (reverse-portrait), which will cause the print filters to generate an "upside-down" image so that the label comes out looking right-side up.

So while I agree having the output the old way can be useful, we need to make sure the output is consistent between drivers and printers.

frazhome:
Thanks a lot for the clarification!
How do I setup this as default setting and is this visible in the webinterface?

PR #4955, opened 2017-02-01
Update log.c to log ISO time stamps

jschwender:
Enhance the logging time stamp to conform with RFC 3339, ISO 8601.

### DIFF
--- a/scheduler/log.c
+++ b/scheduler/log.c
@@ -325,23 +325,7 @@ cupsdGetDateTime(struct timeval *t,	/* I - Time value or NULL for current */
   static struct timeval	last_time = { 0, 0 };
 	    				/* Last time we formatted */
   static char		s[1024];	/* Date/time string */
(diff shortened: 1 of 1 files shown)


michaelrsweet:
CUPS uses Apache Common Log Format, and so uses its date/time format.

Given that we have no desire to change the internally-supported log format and most of the subsequent development in this area is with the system logging daemons, we won't be taking this pull request. Sorry.

jschwender:
Thanks Michael for your response. I understand, this is reasonable. My idea is probaly short-sighted.
Would you consider it if it was alternatively configurable by cupsd.conf as W3C extended format?
Any concerns with that?

michaelrsweet:
Yes, we are not adding more file logging to cupsd. ASL, syslog, systemd logging, etc. all provide support for ISO date/time stamps, and those are the long-term logging solution for CUPS.

PR #4977, opened 2017-03-10
Enabling CUPS to work with recent configuration from /etc/resolv.conf without restart

zdohnal:
Hi Mike,

Our customer reported problem with CUPS about it is not using recent configuration from /etc/resolv.conf without CUPS' restart. It is issue with glibc, but it is known issue for upstream and they are working on solution, but it will not be solved in near future. So I tried to create workaround for CUPS to make it use recent nameserver's IP from /etc/resolv.conf. 
This patch is based on comparing modification times (acquired by stat()) and if they differ, res_init() is called. All of these are encapsulated in function httpCheckResolv(), which is called before all functions, which works with nameservers.
 When I looked into code I saw you are using res_init() when hostname/IP resolution went wrong, so I am checking need_res_init value and if it is set to 1 and my function already reloaded resolver, it sets need_res_init to 0.
I know that using inotify will be probably more complex solution, but if it is glibc issue to begin with, I thought the stat() function will be sufficient as temporary workaround for CUPS.
 I tested this patch on RHEL7 with CUPS-1.6.3 and it worked. So do you want to add this workaround to CUPS, Mike? I know there are some imperfections in this patch (like not checking return status in some places - because I didn't get how you deal with errors there), but I am willing to help with correcting it by your advices.

### DIFF
--- a/cups/auth.c
+++ b/cups/auth.c
@@ -516,6 +516,17 @@ cups_gss_getname(
   DEBUG_printf(("7cups_gss_getname(http=%p, service_name=\"%s\")", http,
                 service_name));
 
+#ifdef HAVE_RES_INIT
+ /*
+  * Check if /etc/resolv.conf is modified.
+  * If so, reload resolver.
+  */
+
+  http_resolv_check_t status;
+
+  status = httpCheckResolv();
+#endif /* HAVE_RES_INIT */
+
 
  /*
   * Get the hostname...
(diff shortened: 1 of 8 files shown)


michaelrsweet:
Sorry, but I don't want to include this in CUPS. The fix belongs below CUPS (and the simplest would be to patch glibc to do the same stat check) since it affects all long-running processes, not just cupsd.

PR #4989, opened 2017-04-11
Fix cupsEnumDests to finish when receiving the ALL_FOR_NOW avahi event

antlarr:
cupsEnumDests listens for avahi answers for `msec` milliseconds (where
`msec` is given as a parameter to cupsEnumDests). avahi sends an
ALL_FOR_NOW event [1] when "more records will probably not show up in
the near future". This means that from the reception of ALL_FOR_NOW
to the timeout triggering, cupsEnumDests basically does nothing but
block the caller. Also, since the caller can't know how long it will
take cupsEnumDests to receive all information from avahi, this means
developers are encouraged to use "big enough" values for `msec`,
thus blocking their calling threads unnecessarily.

This patch makes cupsEnumDests react to the ALL_FOR_NOW event and
finish as soon as all avahi caches have been read and all static
servers have been queried.

No behavior is changed when HAVE_DNSSD is defined.

[1] http://www.avahi.org/doxygen/html/defs_8h.html#af7ff3b95259b3441a282b87d82eebd87a430e1134047305e7b892e784cf72f815

### DIFF
--- a/cups/dest.c
+++ b/cups/dest.c
@@ -85,6 +85,7 @@ typedef struct _cups_dnssd_data_s	/* Enumeration data */
   AvahiSimplePoll	*simple_poll;	/* Polling interface */
   AvahiClient		*client;	/* Client information */
   int			got_data;	/* Did we get data? */
+  int			remaining_browsers; /* Remaining running avahi browsers */
 #  endif /* HAVE_DNSSD */
   cups_dest_cb_t	cb;		/* Callback */
   void			*user_data;	/* User data pointer */
@@ -1029,6 +1030,7 @@ cupsEnumDests(
   data.cb        = cb;
   data.user_data = user_data;
   data.devices   = cupsArrayNew3((cups_array_func_t)cups_dnssd_compare_devices, NULL, NULL, 0, NULL, (cups_afree_func_t)cups_dnssd_free_device);
+  data.remaining_browsers = 1;
 
 #  ifdef HAVE_DNSSD
   if (DNSServiceCreateConnection(&data.main_ref) != kDNSServiceErr_NoError)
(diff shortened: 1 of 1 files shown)


michaelrsweet:
mDNSResponder has a similar mechanism, however in practice it has not proven reliable since all it means is that there are no more responses queued up... Will think about how we can make this better...

michaelrsweet:
OK, I have a better (generic) fix for this that uses some of the tricks from the dnssd backend to return early when we know we have all responses. Doing some more testing and will push a change once I'm satisfied.

michaelrsweet:
Sorry, no, I'm just working directly on master for this change. But it shouldn't be too much longer before I push it to Github...

michaelrsweet:
[master a2187a634] Update cupsEnumDests implementation to return early if all printers have been discovered (Issue #4989)

michaelrsweet:
Please verify the fix in Github resolves the problems you were having.

antlarr:
I added a couple of comments to your commit. Do you want me to modify my PR to fix those? Or do you prefer to work on them yourself?

michaelrsweet:
@antlarr I'm doing follow-up testing on a few Linux distros now (first pass mainly tested macOS) and had found the issues you have reported in the comments to the commit. I have some more changes for Avahi coming...

michaelrsweet:
OK, just pushed more changes that seem to make Ubuntu happy.

antlarr:
Perfect, with a2187a63425a3d6c05de1e1cbf8c26fd39a1aced, 657c5b5f91e6d5120c4ad7b118cf9098dd27f03d and 3fae3b337df0be1a766857be741173d8a9915da7 now it works fine.

Thanks, I'll include those three commits in openSUSE Leap/Tumbleweed and SUSE Linux Enterprise. I'll close this pull request if you think that's ok now.

michaelrsweet:
Yes, I'll probably keep testing on our other supported platforms (Windows in particular) but you should be good to go on Linux.

PR #5029, opened 2017-06-11
locale: introduce cups_zh_CN.po

MingcongBai:
Not sure if there's anything else to do... Haven't tested it out yet.


MingcongBai:
@michaelrsweet Thanks so much!

PR #5066, opened 2017-07-22
Update french translations

sblondon:
This patch provides an update of cups_fr.po : it adds several strings and fixes few of the previous ones.

There are still a lot of untranslated strings. However I think it's a good first step.
 
Do you think it's mergeable? Do you want more french translations?


michaelrsweet:
Will review and merge if everything looks OK.

michaelrsweet:
Looks good visually, will merge and then correct any errors that checkpo shows...

sblondon:
Thanks!

PR #5098, opened 2017-08-28
Disable port comparison in _cupsConnect() when connection type is AF_UNIX

Cistoge:
Hello,

I could not print with cups 2.2.1, at least not with the command line tools, when policy was set to authenticated. 

First, I found that the cause was the comparison of cg->ipp_port and httpAddrPort(cg->http->hostaddr)) in _cupsConnect() in request.c, because when connection type is AF_UNIX, the first is 631 and the second is 0. That causes the connection to be closed and a new one to be established immediately after, but the new connection does not have the authentication information of the previous connection.

Then, I found that this was introduced when httpAddrPort() in http-addr.c was changed so that it returns 0 instead of 631 in the case when connection type is AF_UNIX. As it makes sense to return 0 as port number in that case, I did not change httpAddrPort(), but _cupsConnect() in request.c, so that the port comparison between cg->ipp_port and httpAddrPort(cg->http->hostaddr)) only takes place in non-AF_UNIX connections.

I attached a patch and would like it to be reviewed.

Regards
  Christoph

[no_port_comparison_with_unix_socket.patch.txt](https://github.com/apple/cups/files/1257147/no_port_comparison_with_unix_socket.patch.txt)


michaelrsweet:
Pushed a similar fix:

[master 9449dd948] The CUPS library did not reuse domain sockets (Issue #5098)

Issue #5041, opened 2017-06-30
SystemGroup cannot handle groups with spaces


bluefoxicy:
In Cups 2.2.0, the following line works in cups-files.conf:

    SystemGroup "a group" sys root

In Cups 2.2.2, `cupsd -t` reports that the file contains errors for any variant:

    SystemGroup "a group" sys root
    SystemGroup 'a group' sys root
    SystemGroup a\ group sys root

michaelrsweet:
First, spaces are not generally supported in UNIX group names - that it ever worked is blind luck...

Second, the code that parses the system groups hasn't changed in over 10 years and *does* support quoted but not backslash escaped values. Both of the quoted SystemGroup definitions should work, as long as "a group" exists as a group. But note that groups in /etc/group cannot contain spaces on Linux and macOS (at least).

Also, what is the exact error message you are seeing?

bluefoxicy:
> First, spaces are not generally supported in UNIX group names

Too bad.  UNIX supports NIS, LDAP, and other information sources which _do_ support spaces, and doesn't generally filter them to replace spaces with another character.  This is one of the few common questions that pops up on forums and mailing lists about configuring CUPS, as much as there are any common questions about configuring CUPS.

> Second, the code that parses the system groups hasn't changed in over 10 years and does support quoted but not backslash escaped values.

Then I'll need to re-examine my set-up.  It works in Ubuntu 16.10 (Cups 2.2.0) but not 17.04 (2.2.2); perhaps it is a distribution issue, or a configuration error on my part.  I'll find out next week.

> Also, what is the exact error message you are seeing?

`cupsd` just exits with code 2, no logs, and no output.

`cupsd -t` says `/etc/cups/cups-files.conf contains errors` and exits.  It works if I remove the quoted group from `cups-files.conf`.

michaelrsweet:
OK, it appears the error messages are going to the system log (since no log file has been opened yet), rather then the standard error... The quotes are being maintained as part of the lookup.

michaelrsweet:
OK, I've committed changes that force logging to stderr when "-t" is used, and improved the logging of SystemGroup errors.

That said, the code (still) does the right thing with quoted strings, so more than likely the issue is that getgrnam() is not finding the named group.

OdyX:
This change makes builds on systems where a system group doesn't exist fail in the testsuite, e.g.:                 ```
--with-system-groups=lpadmin
```

```
FAIL: 35 error messages, expected 33.
[...]

PR #5099, opened 2017-08-28
Tests: error_log entries counting is off by two 'Unknown default SystemGroup' occurences

OdyX:
This regression was introduced by the fix for https://github.com/apple/cups/issues/5041 which makes **builds** with `--with-system-groups=lpadmin` in environments where that same system group doesn't exist fail in the testsuite with the following error:

```
FAIL: 35 error messages, expected 33.
E [28/Aug/2017:13:46:06.572986 +0000] Unknown default SystemGroup "lpadmin".
…
```
The `error_log` error counting is off by two additionnal error messages as above.

This patch is how we fixed it in Debian, adressing specifically this problem only. There's certainly a better way to fix this than omitting that line in the `error_log` lines counting.

By the way; these are all the exceptions we take into account to only count '33' errors consistently:
```
count=`$GREP '^E ' $BASE/log/error_log | \
       $GREP -v '(usb) crashed on signal 11' | \
       $GREP -v '(dnssd) stopped with status 1' | \
       $GREP -v 'loadFile failed: temp file: not a PDF file' | \
       $GREP -v 'Failed to connect to system bus' | \
       $GREP -v -E 'Unable to open listen socket for address .* Address family not supported by protocol.' | \
       $GREP -v 'Unable to write uncompressed print data: Broken pipe' | \
       $GREP -v 'Unknown default SystemGroup' | \
       wc -l | awk '{print $1}'`
```

### DIFF
--- a/test/run-stp-tests.sh
+++ b/test/run-stp-tests.sh
@@ -996,7 +996,9 @@ else
 fi
 
 # Error log messages
-count=`$GREP '^E ' $BASE/log/error_log | wc -l | awk '{print $1}'`
+count=`$GREP '^E ' $BASE/log/error_log | \
+       $GREP -v 'Unknown default SystemGroup' | \
+       wc -l | awk '{print $1}'`
 if test $count != 33; then
 	echo "FAIL: $count error messages, expected 33."
 	$GREP '^E ' $BASE/log/error_log


michaelrsweet:
Not sure why you are seeing extra messages, but will investigate...

michaelrsweet:
OK, well for now filtering the error messages seems to be the least invasive way to avoid this:

[master 979804902] Fix "make check" for restricted environments (Issue #5099)

PR #5101, opened 2017-08-31
Fix authorization check for kerberos local connections

kernevil:
When a client connects through local socket, it is always authenticated
using Basic and socket peer credentials regardless the authentication
method configured. The authorization function should take this into
account and skip the authentication method check for AF_LOCAL connections.

Signed-off-by: Samuel Cabrero <scabrero@suse.de>

### DIFF
--- a/scheduler/auth.c
+++ b/scheduler/auth.c
@@ -1694,6 +1694,9 @@ cupsdIsAuthorized(cupsd_client_t *con,	/* I - Connection */
 
 
     if (con->type != type && type != CUPSD_AUTH_NONE &&
+#ifdef AF_LOCAL
+        (_httpAddrFamily(con->http.hostaddr) != AF_LOCAL) &&
+#endif /* AF_LOCAL */
 #ifdef HAVE_GSSAPI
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Need to do a security review - at first blush this isn't the right fix (we should be using Kerberos for local connections) but will check...

sriemer:
The problem exists together with Samba. Kerberos auth is used for the CUPS web interface and Kerberos is used for the Windows clients. When pausing/resuming printers with Windows Explorer, Samba has to authenticate to CUPS (again).

michaelrsweet:
@sriemer Basically, you are asking for the local user account (whatever Samba is using) to be able to bypass the Kerberos authentication requirement. But a Samba user may not have a corresponding local user account, right? This could end up being "nobody" or "root" or some other account, right?

Unless we know that the Samba user exists as a local user account, this doesn't look like the correct fix (whatever that fix should be).

michaelrsweet:
and after going through the auth code again (as part of a HTTP Digest modernization effort), it looks like neither cupsd nor libcups will use peer credentials if Kerberos (Negotiate) is enabled.

Closing this PR. Please open a new one if you come up with a solution...

PR #5103, opened 2017-09-01
Add a USB quirk rule for the Kyocera Ecosys P6130cdn (Issue #5102)

### LINKED ISSUE #5102: Kyocera Ecosys P6130cdn needs usb-no-reattach-default=true
When printing two documents very quickly after one another on the Kyocera Ecosys P6130cdn connected by USB, the output of the second printout is incorrect.

The behavor is basically as already described in issue #4900 for the model Kyocera Ecosys P6026cdn, except for the exact output now being different:

~~~
Error Name:          /undefined
Offending Command:   umCopies

Operand Stack:
~~~

It can be fixed by adding the following lines to the USB quirks file (`/usr/share/cups/usb/org.cups.usb-quirks` in Ubuntu):

~~~
# Kyocera Ecosys P6130cdn
0x0482 0x0677 no-reattach
~~~

Can this be added to upstream CUPS?

michaelweghorn:
This adds the respective USB quirk entry needed to fix #5102.

### DIFF
--- a/backend/org.cups.usb-quirks
+++ b/backend/org.cups.usb-quirks
@@ -251,5 +251,8 @@
 # Kyocera Ecosys P6026cdn (Issue #4900)
 0x0482 0x063f no-reattach
 
+# Kyocera Ecosys P6130cdn (Issue #5102)
+0x0482 0x0677 no-reattach
+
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Thanks!

PR #5122, opened 2017-09-25
configure: improve interpreter detection

rohieb:
Currently, when configure is called with `--with-java` or `--with-java=`, auto-detection is performed. When called with `--without-java`, `HAVE_JAVA` is still being defined. This is unfortunate when cross-compiling for embedded systems, we would end up with Java on the host being auto-detected (which is not what we have on the target), or with a wrong `HAVE_JAVA` define.

This implements a way to explicitely disable scripting support for all supported languages with ``without-java` (or the respective options for Perl, PHP and Python)

While we're at it, make the detection a bit more verbose, and do some cleanup.

### DIFF
--- a/config-scripts/cups-scripting.m4
+++ b/config-scripts/cups-scripting.m4
@@ -16,6 +16,7 @@ AC_ARG_WITH(java, [  --with-java             set Java interpreter for web interf
 	CUPS_JAVA="$withval",
 	CUPS_JAVA="auto")
 
+AC_MSG_CHECKING([for Java interpreter])
 if test "x$CUPS_JAVA" = xauto; then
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Ultimately I think I want to eliminate all of the third-party scripting support - only Java "needs" special handing as the others support the #! interpreter mechanism...

rohieb:
Ah ok, yes, that seems reasonable. Feel free to only pick the Java part from this PR.

michaelrsweet:
I pushed a slightly different change:

[master b889560c0] Fixed the script interpreter detection in the configure script (Issue #5122)

PR #5123, opened 2017-09-26
socket backend doesn't fail when printer is disconnected

zdohnal:
Hi Mike,

we ran several test and found out that socket backend doesn't return failure if printer is disconnected or powered off. httpAddrConnect() function returns ETIMEDOUT or ENOTCONN in these cases and code doesn't catch them. Created patch catches them and returns failure.
Is it acceptable behavior for socket backend?

### DIFF
--- a/backend/socket.c
+++ b/backend/socket.c
@@ -331,7 +331,8 @@ main(int  argc,				/* I - Number of command-line arguments (6 or 7) */
       fprintf(stderr, "DEBUG: Connection error: %s\n", strerror(error));
 
       if (error == ECONNREFUSED || error == EHOSTDOWN ||
-          error == EHOSTUNREACH)
+          error == EHOSTUNREACH || error == ENOTCONN ||
+          error == ETIMEDOUT)
       {
         if (contimeout && (time(NULL) - start_time) > contimeout)
 	{
@@ -353,6 +354,16 @@ main(int  argc,				/* I - Number of command-line arguments (6 or 7) */
 				   _("The printer is unreachable at this "
 				     "time."));
 	      break;
+
+    case ETIMEDOUT :
+        _cupsLangPrintFilter(stderr, "WARNING",
+           _("The printer connection timed out."));
+        break;
+
+    case ENOTCONN :
+        _cupsLangPrintFilter(stderr, "WARNING",
+            _("Unable to connect to the printer."));
+        break;
 
 	  case ECONNREFUSED :
 	  default :


michaelrsweet:
We should do something; reviewing...

jsmeix:
Please, be very careful before you let a backend
exit with non-zero exit status because that can
by default disable the whole print queue.

In particular when a printer is disconnected or powered off
it may re-appear after a short time because when a print job
does not print out as expected the usual user bahaviour is
to disconnect or power off the printer (regardless what one
may tell users what "the right way" would be in such cases)
users will first and foremost disconnect or power off the printer
in case of any issue.

In the past there were tons of annoyed user complaints
about CUPS because of the default backend behaviour
to exit too early in case of issues that look like a final error
but were actually only temporary issues, e.g. issue like
https://bugzilla.opensuse.org/show_bug.cgi?id=337665
https://bugzilla.opensuse.org/show_bug.cgi?id=338605
and
https://bugzilla.opensuse.org/show_bug.cgi?id=337794

Those tons of annoyed user complaints even lead to
special workarounds like the 'beh' backend error handler
which cause (as any workaround) its own special problems
(e.g. the 'hp' backend from HPLIP did not work with 'beh').

Because a printer may re-appear after a short time
after it was disconnected or powered off, a backend
should wait a reasonable long time (better a bit longer
than a bit too short) for the printer to-re-appear and
if the printer re-appears then "do the right thing" which
is not so easy:

I think when a printer gets "lost" while the backend runs
but the backend did not yet send a single byte of data
to the printer then the backend can wait a very long time
for the printer to re-appear and if it re-appears the
backend can re-start its communication with the printer.

If the printer does not re-appear after a long timeout time
I think this is still no sufficient reason to disable the queue.
Perhaps the user has a laptop and moved into another
network and comes back three days later (e.g. on Monday
after the weekend). Then his queue should not have been
disabled. But I think the user is then probably no longer
interested in his old print job from three days ago so that
the old job does not need to be reprinted.

I think when a printer gets "lost" while the backend is
already sending data to it and later it re-appears (while
the same backend process is waiting for the printer
to re-appear), there is in general no way how to
correctly resume such an interrupted print job.
Accordingly I think the backend should exit with
something like CUPS_BACKEND_HOLD or
CUPS_BACKEND_CANCEL or
CUPS_BACKEND_RETRY[_CURRENT].

I think when a printer gets "lost" while the backend
is sending data to it and it does not re-appears during
the timeout time while the same backend process is
waiting for the printer to re-appear, the backend must
not exit with CUPS_BACKEND_STOP or
CUPS_BACKEND_FAILED because I think this is
not a sufficient reason to stop the whole print queue.
I think in this case CUPS_BACKEND_CANCEL
is the only right one because when a printer got "lost"
for a long time it means the user intended that
"the job was canceled at the printer"
(cf. "man 7 backend").

Cf. what I had written a loger time ago when backends
exit too early in case of temporary issues in
https://bugzilla.opensuse.org/show_bug.cgi?id=337794#c16
therein in particular what I had written at that time that the
"backends should be more robust against possibly transient errors"
(where some details are nowadays probably somewhat outdated).

zdohnal:
@jsmeix  Ok, I'll change the patch to reflect your advices - if ENOTCONN or ETIMEDOUT came, socket backend will return CUPS_BACKEND_CANCEL. Is it acceptable?

michaelrsweet:
The correct behavior will be to retry the connection unless the job was submitted to a class or the job has been canceled. Neither error is fatal.

zdohnal:
Truth is we are writing new failover backend, which I want to show you or Till because of possible adding to your project when backend will be complete. This failover backend works in similar way like classes, but user can prioritize queues, which influences to which queues job arrives first. For implementing this behavior, I need socket backend to end and send info to failover backend about success/failure, so other queue can be tried.
If I understand correctly, there are two advices:
1) do not use CUPS_BACKEND_FAILED
2) backend needs to retry connection
So there is suggestion - would it be acceptable that when errno will be ETIMEDOUT or ENOTCONN, socket backend will retry the connection until contimeout expires and returns CUPS_BACKEND_CANCEL or new error code CUPS_BACKEND_TIMEDOUT (used only for this)?

Something like this:
 if (error == ECONNREFUSED || error == EHOSTDOWN ||
     error == EHOSTUNREACH || error == ENOTCONN || error == ETIMEDOUT)
{
  if (contimeout && (time(NULL) - start_time) > contimeout)
  {
    _cupsLangPrintFilter(stderr, "ERROR",
                         _("The printer is not responding."));
    if (error == ENOTCONN || error == ETIMEDOUT)
      return (CUPS_BACKEND_CANCEL);
    else
      return (CUPS_BACKEND_FAILED);
  }

michaelrsweet:
No, we don't want to add more exit codes or change behavior here. Nor will we incorporate yet another backend to implement different prioritization or queues, etc.

The long-term goal of the CUPS project is to eliminate PPD files and printer drivers, and only use IPP to communicate with printers and servers. Ultimately that will result in a vast simplification of the server code which will allow us to offer things like prioritized queues without worrying that some printer driver or backend will break things.

zdohnal:
Ok, thank you for answer. Closing this pull request.

michaelrsweet:
I actually pushed the following change for this:

> [master 4cf66fef4] The network backends now retry on more error conditions (Issue #5123)

Basically I added ETIMEDOUT and ENOTCONN to the list of networking errors we potentially retry.

zdohnal:
Thank you Mike :) . But I am not able to see the commit you mentioned. Is it some error in git?

michaelrsweet:
Sorry, push must have failed yesterday. Pushed now...

zdohnal:
No problem, thank you Mike.

Issue #5135, opened 2017-10-03
cupsGetDests2() lists local printers even if remote server specified


MarekKasik:
cupsGetDests2 function lists local printers even if it get "http_t *" parameter which points to a remote server. It seems to me that this is caused by the fix of the issue #4993. The new implementation does not use the "http" parameter at all.
Was it intentional?

michaelrsweet:
The pointer should still be used, but keep in mind that support for non-local CUPS servers has been deprecated since 2.0 due to sandboxing and other security "features" preventing clients from accessing network printers directly - that's one of the reasons we added the new CUPS-Create-Local-Printer operation so that cupsd did all of the communications for you...

Investigating...

michaelrsweet:
OK, I see how cupsGetDests2 is calling cupsEnumDests, which doesn't have a http_t * argument...  Will fix this for 2.2.5...

zdohnal:
I'll just want to mention this issue breaks searching for printers in gnome (in control-center package).

zdohnal:
I created patch, which does the trick (but it changes cupsEnumDests() and cupsEnumDestsBlock() API - it adds new parameter http_t *http - so I am not sure if it is right way to do it). https://github.com/apple/cups/pull/5136

michaelrsweet:
[master 9554d4e74] `cupsGetDests2` was not using the supplied HTTP connection (Issue #5135)

zdohnal:
Mike, I think commit didn't propagate to github master.

MarekKasik:
Thank you for the fix.

zdohnal:
Oh, nice, I can see it now - thank you for the fix Mike!

PR #5136, opened 2017-10-04
Proposed patch for issue #5135

zdohnal:
More info in issue #5135

### DIFF
--- a/cups/cups.h
+++ b/cups/cups.h
@@ -523,12 +523,12 @@ extern ipp_status_t	cupsCreateDestJob(http_t *http, cups_dest_t *dest,
 					  cups_dinfo_t *info, int *job_id,
 					  const char *title, int num_options,
 			                  cups_option_t *options) _CUPS_API_1_6;
-extern int		cupsEnumDests(unsigned flags, int msec, int *cancel,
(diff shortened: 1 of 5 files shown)


zdohnal:
Only other thing to do is create something like cupsEnumDests2(), which is used for non default http connection - but it will need to extract code from cupsEnumDests(), where http needs to be used and then call cupsEnumDests(), but I do not see any simple way how to do it. For now I use duplicate of cupsEnumDests, which uses http.

michaelrsweet:
We won't change public API in a way that breaks source or binary compatibility, and we have no plans of updating cupsEnumDests to take a http_t pointer argument since the use of remote CUPS servers has been deprecated for a long time due to incompatibilities with sandboxing on macOS and the various security features on the different Linux distros.

zdohnal:
If I create the patch without breaking public API, would it be acceptable?

michaelrsweet:
We will come up with a patch shortly. Please be patient.

zdohnal:
I am sorry, I didn't want to sound impatient, I just wanted to help you with solution (I have patch without breaking public API now, I can create pull request anytime). To sum it up, I apologize.

PR #5147, opened 2017-10-17
 dnssd-discovered (by avahi) device URIs not usable

zdohnal:
Hi Mike,
I was checking old patches we have and I've found this one - apparently dnssd devices discovered by avahi doesn't work, because current code uses addresses from list returned from httpAddrGetList(), not address from Avahi.
Patch was created by Tim Waugh in the past and there is bug report from the past https://bugzilla.redhat.com/show_bug.cgi?id=948288 . Code in these parts of http-support.c hasn't changed since 2012, report was created in 2013, I think the issue is there still.

### DIFF
--- a/cups/http-support.c
+++ b/cups/http-support.c
@@ -2346,7 +2346,7 @@ http_resolve_cb(
     const char             *type,	/* I - Registration type */
     const char             *domain,	/* I - Domain (unused) */
     const char             *hostTarget,	/* I - Hostname */
-    const AvahiAddress     *address,	/* I - Address (unused) */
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Ok, if we can’t resolve a .local host name, that would be a serious distribution or avahi bug. Certainly on Ubuntu the resolution of .local names via getnameinfo works, and you MUST support multiple addresses for mDNS or face serious interop issues (which is why httpAddrConnect tries connections to all of the resolved addresses)

This isn’t the correct fix.

Issue #5145, opened 2017-10-16
SSLOptions not working at all, cups 2.2.5 breaks CBC TLS suites even without DenyCBC


hardfalcon:
The `DenyCBC` parameter introduced in CUPS 2.2.5 for the `SSLOptions` directive doesn't work at all, it is simply always on because the corresponding logic in [cups/tls-gnutls.c](https://github.com/apple/cups/blob/v2.2.5/cups/tls-gnutls.c) is broken in at least two ways.

First, [the if() clause on line 1523](https://github.com/apple/cups/blob/v2.2.5/cups/tls-gnutls.c#L1523) triggers when the `_HTTP_TLS_DENY_CBC` bit in the `tls_options` variable is _not_ set, which is the exact opposite of what it is supposed to do.

Second, the `tls_options` variable [is always zero](https://github.com/apple/cups/blob/v2.2.5/cups/tls-gnutls.c#L1229) on my machine (Archlinux) regardless of what parameters I pass to the `SSLOptions` directive. Unfortunately, I haven't been able to find the culprit yet for this.

In combination, this leads to the DenyCBC behaviour being always applied regardless of the actual configuration in cupsd.conf, and instead [produces](https://github.com/apple/cups/blob/v2.2.5/cups/tls-gnutls.c#L1506) a `priority_string` of `NORMAL:+VERS-TLS-ALL:-VERS-SSL3.0:-ARCFOUR-128:!ANON-DH:!AES-128-CBC:!AES-256-CBC:!CAMELLIA-128-CBC:!CAMELLIA-256-CBC:!3DES-CBC`.

This effectively kicks out all ciphersuites that could be supported by non-TLS1.2 clients (only AES-GCM, AES-CCM and CHACHA20-POLY1305 suites remain). The result is that Windows 7 clients cannot print anymore. For some bizarre reason, the Windows printing system only uses TLS 1.0 and SSL 3.0 by default, and you have to [fiddle with the registry](https://support.microsoft.com/en-us/help/3140245/update-to-enable-tls-1-1-and-tls-1-2-as-a-default-secure-protocols-in) to get support for TLS 1.1 and TLS 1.2.

[...]

PR #5149, opened 2017-10-19
Clean up implementation for SSLOptions

hardfalcon:
This cleans up the implementation for the `SSLOptions` configuration directive:

- Fixed the logic for `DenyCBC` on Linux and other gnutls platforms.
- Fixed the logic for `_httpTLSSetOptions()` in `cups/usersys.c` to prevent unsolicited overwriting of the `SSLOptions` read from `cupsd.conf`.
- The `options` / `tls_options` variables got changed into unsigned integers since we are using them as bit maps and not as "normal" numbers.
- The `_HTTP_TLS_*` constants got shifted by one bit to accomodate the new constant `_HTTP_TLS_UNCHANGED`, which means that the values changed for all of these constants. This is needed so we can distinguish between "explicitly use the default settings" and "don't change/overwrite the existing configuration".
- In some places, numerical values were replaced with the corresponding `_HTTP_TLS_*` constants.
- Added explicit warnings about the `AllowDH`, `AllowRC4` and `AllowSSL3` parameters to the manual.

### DIFF
--- a/cups/http-private.h
+++ b/cups/http-private.h
@@ -179,13 +179,14 @@ extern "C" {
 #define _HTTP_RESOLVE_FQDN	2	/* Resolve to a FQDN */
 #define _HTTP_RESOLVE_FAXOUT	4	/* Resolve FaxOut service? */
 
-#define _HTTP_TLS_NONE		0	/* No TLS options */
-#define _HTTP_TLS_ALLOW_RC4	1	/* Allow RC4 cipher suites */
-#define _HTTP_TLS_ALLOW_SSL3	2	/* Allow SSL 3.0 */
-#define _HTTP_TLS_ALLOW_DH	4	/* Allow DH/DHE key negotiation */
-#define _HTTP_TLS_DENY_TLS10	16	/* Deny TLS 1.0 */
-#define _HTTP_TLS_DENY_CBC	32	/* Deny CBC cipher suites */
-#define _HTTP_TLS_ONLY_TLS10    64      /* Only use TLS 1.0 */
+#define _HTTP_TLS_UNCHANGED     0   /* Don't change TLS options */
+#define _HTTP_TLS_NONE          1   /* No TLS options */
+#define _HTTP_TLS_ALLOW_RC4     2   /* Allow RC4 cipher suites */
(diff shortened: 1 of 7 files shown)


michaelrsweet:
OK, so we can't take this pull request as-is, particularly the wording changes for the Allow* value documentation.

hardfalcon:
Any proposition for a better wording? Especially the `AllowDH` parameter is _really_ dangerous, since it effectively enables MITM attacks without any computational cost. The `ANON-DH` ciphersuites it enables in gnutls correspond to the DH subset of the `aNULL` ciphersuites in OpenSSL - which means they do not offer any authentication whatsoever.

I'm still wondering how `ANON-DH` could make it into CUPS at all - even if there were printers which required such broken ciphersuites, it would still be better to access them via plain HTTP/IPP instead of globally forcing the whole CUPS server to offer these ciphersuites.

hardfalcon:
I've just cross-checked with `cups/tls-darwin.c` - on macOS, all anonymous ciphersuites are explicitly disabled with even a [remark](https://github.com/apple/cups/blob/v2.2.5/cups/tls-darwin.c#L1257) in the code that they should never be enabled because they're insecure.

It seems that gnutls simply doesn't support/implement static DH key exchanges at all, so it's simply impossible to implement the `AllowDH` parameter on gnutls platforms. I'd propose changing it into a no-op and updating the manual accordingly.

michaelrsweet:
The wording we have is neutral about the security implications of the Allow* options because a) they are disabled by default for a reason, which a simple Google search will show, and b) typically if you have to use an Allow* option it is because you have no other choice. Calling things horrible or insecure is unnecessary.

As for the AllowDH option, if indeed GNU TLS no longer implements any DH cipher suites except for the NULL ones, then we should probably not support it there. Additionally I'll make sure we unconditionally exclude all NULL suites.

hardfalcon:
Wording: I'd prefer to explicitly warn users about insecure options, but it's your project, so your call, your rules. I've updated the PR accordingly.

AllowDH: I've turned AllowDH into a no-op on gnutls and updated the manual accordingly.

I've also added notes about the platforms on which the specific SSLOptions are supported.

michaelrsweet:
[master 02c88e673] Fix cipher suite selection with GNU TLS (Issue #5145)

Issue #4991, opened 2017-04-24
Total page counter is inconsistent in the page_log


akorobkin:
When printing with CUPS 2.2.3, cups-filters 1.13.4, the number of individual pages in the log are correct, but the "total" line is almost always erroneous. Sometimes it is 0, which is strange. Sometimes it appears twice after printing one file. 

Excerpt from the log:
```
dalek-color korobkin 58 [24/Apr/2017:08:58:38 -0700] 1 1 - 100.106.90.151 AIUWATPoster.pdf Tabloid -
dalek-color korobkin 58 [24/Apr/2017:09:00:54 -0700] total 7 - 100.106.90.151 AIUWATPoster.pdf Tabloid -
dalek-color korobkin 59 [24/Apr/2017:09:00:54 -0700] 1 1 - 100.106.90.151 tillA4-A3-A4-3-pdftopdf.pdf Letter -
dalek-color korobkin 59 [24/Apr/2017:09:00:54 -0700] 2 1 - 100.106.90.151 tillA4-A3-A4-3-pdftopdf.pdf Letter -
dalek-color korobkin 59 [24/Apr/2017:09:00:54 -0700] 3 1 - 100.106.90.151 tillA4-A3-A4-3-pdftopdf.pdf Letter -
dalek-color korobkin 60 [24/Apr/2017:09:01:15 -0700] 1 2 - 100.106.90.151 3-pages-double-sided-stapled.pdf Letter -
dalek-color korobkin 60 [24/Apr/2017:09:01:15 -0700] 2 2 - 100.106.90.151 3-pages-double-sided-stapled.pdf Letter -
dalek-color korobkin 60 [24/Apr/2017:09:01:15 -0700] 3 2 - 100.106.90.151 3-pages-double-sided-stapled.pdf Letter -
dalek-color korobkin 60 [24/Apr/2017:09:01:19 -0700] total 0 - 100.106.90.151 3-pages-double-sided-stapled.pdf Letter -
dalek-color korobkin 61 [24/Apr/2017:09:01:19 -0700] 1 2 - 100.106.90.151 7-pages-double-sided-short-edge.pdf Letter -
dalek-color korobkin 61 [24/Apr/2017:09:01:19 -0700] 2 2 - 100.106.90.151 7-pages-double-sided-short-edge.pdf Letter -
[...]

PR #5153, opened 2017-10-25
snmp page counting is unreliable - give way to disable it in PPD file

zdohnal:
Hi Mike,

our customer found out that snmp page counting is unreliable for some printers (printers seems to send end of printing before life counter in printer snmp agent got updated - this behavior resulted in bad page count in cups web ui), so we implemented the way to turn off page counting by snmp. User simply add "cupsSNMPPages: False" into printer ppd file and cups then uses counting by 'PAGE X 1' messages.
I understand CUPS project is on its way to drop .ppd functionality, but IMHO it is good to offer opportunity for these, which still uses ppd files and they complain about snmp page counting, until better solution is implemented.

### DIFF
--- a/backend/snmp-supplies.c
+++ b/backend/snmp-supplies.c
@@ -83,6 +83,8 @@ static backend_supplies_t supplies[CUPS_MAX_SUPPLIES];
 					/* Supply information */
 static int		supply_state = -1;
 					/* Supply state info */
+static int              use_snmp_pages = 1;
+                                        /* Report pages used or not */
 
 static const int	hrDeviceDescr[] =
 			{ CUPS_OID_hrDeviceDescr, 1, -1 };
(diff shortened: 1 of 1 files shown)


michaelrsweet:
@zdohnal How is this different from the *cupsSNMPSupplies keyword that does the same thing (plus supply level information)?

I really don't want to add yet another switch for controlling a legacy protocol.

zdohnal:
It only switches off SNMP page counting - supply level info is still shipped by SNMP.

michaelrsweet:
Still not convinced, but will hold this possibly for a future release.

zdohnal:
I can possibly rewrite it into similar way as cupsSNMPSupplies are applied - like command line option, if it helps. But it didn't look like as such serious issue with high impact to have extra command line option - but sure, it will look better if it can be set by command line. Is it acceptable?

michaelrsweet:
OK, the changes for issue #4991 should mitigate the issues associated with lowball accounting numbers - we now use MAX(printer-page-count,filter-page-count) when reporting the number of pages that were printed.

michaelrsweet:
OK, I'm punting on this pull request since our future focus will be on modern protocols...

PR #5175, opened 2017-11-17
Increase size of max header length to accomodate token authentication

kschlepp:
Some tokens are larger than the current max header value length allows for.

### DIFF
--- a/cups/http.h
+++ b/cups/http.h
@@ -96,7 +96,7 @@ extern "C" {
 #  define HTTP_MAX_URI		1024	/* Max length of URI string */
 #  define HTTP_MAX_HOST		256	/* Max length of hostname string */
 #  define HTTP_MAX_BUFFER	2048	/* Max length of data buffer */
-#  define HTTP_MAX_VALUE	256	/* Max header field value length */
+#  define HTTP_MAX_VALUE	2048	/* Max header field value length */
 
 
 /*


michaelrsweet:
No, aside from increasing the size of the http_t structure 8-fold, this change would break binary compatibility.

What header do you need larger for authentication? We already store the Authorization header in an allocated string, and for CUPS 2.3.x and larger for WWW-Authenticate as well.

kschlepp:
We need to use HTTP_FIELD_AUTHENTICATION. Should we be using HTTP_FIELD_WWW_AUTHENTICATE instead?

michaelrsweet:
HTTP_FIELD_AUTHENTICATION (Authentication: ...) is what you send from the client to the server. That one has been allocated for a very long time (since Kerberos was added) so the 256 byte value limit does not apply.

HTTP_FIELD_WWW_AUTHENTICATE (WWW-Authenticate: ...) is what you send from the server to the client in a challenge (401) response. That field was recently just changed to be allocated for the forthcoming CUPS 2.3 release.

What sort of problems are you having? Are the Authentication: headers sent via libcups getting truncated or ???

kschlepp:
There is some confusion on my team about the changes we needed to make. I'm going to close this.

Sorry about that! Thanks for your help!

PR #5213, opened 2018-01-02
Fix ppdInstallableConflict conflict reporting

albert-astals-cid-kdab:
Without this patch it was taking all constraints into account, not only the ones
regarding the passed option so if you had a conflict somewhere else, every other
single option would be said to be conflicting which doesn't match what the
documentation says this function should do.


michaelrsweet:
Looks reasonable, thanks for the fix!

Issue #5225, opened 2018-01-16
When writing CUPS/PWG/Apple Raster compressed, byte-swapping never happens


tillkamppeter:
In the function `cupsRasterWritePixels()` in `filter/raster.c` (libcupsimage) it was completely overlooked that when the output is compressed (r->compressed) AND byte-swap is needed (r-swapped AND high color depth > 8 bit) the byte-swapping has to done before compression.
I have found this by trying to print in the Adobe RGB color space with 16 bits per pixel (ADOBERGB48) in Apple Raster on my DeskJet 2540. This has given the correct structure but wrong colors on the print out. I have fixed the bug and after that I have gotten correct printouts in all 16-bit-per-pixel color spaces.

tillkamppeter:
Fix in pull request #5226.
I have tested all color depth on the DeskJet 2540 (Apple Raster), 8-bit and 16-bit and they are all working correctly now.
Note: rasterview has also a bug. It displayed the files correctly with this bug ond wrong after the fix.

michaelrsweet:
Not taking PR #5226 but will provide a separate fix later today.

michaelrsweet:
[master 7c91db71b] The `cupsRasterWritePixels` function did not correctly swap bytes for some formats (Issue #5225)

tillkamppeter:
I have done the following tests to confirm that it works:
- Build current snapshot of CUPS and install libcupsimage into the system
- For screen display: Take rasterview 1.5 source, run `make`, then rebuild rasterview to use libcupsimage:
```
g++  -o rasterview RasterDisplay.o RasterView.o main.o -fPIE -pie -lfltk_images -lfltk -lX11 -lcupsimage
```
- Take the high-color-depth-using PPD file `x.ppd` (attached, generated with cups-filters from the DeskJet 2540) and generate URF files running filter chains with the `cupsfilter` utility:
```
cupsfilter -p x.ppd -m image/urf -o ColorModel=Gray /usr/share/cups/data/testprint > g.urf
cupsfilter -p x.ppd -m image/urf -o ColorModel=RGB /usr/share/cups/data/testprint > s.urf
cupsfilter -p x.ppd -m image/urf -o ColorModel=AdobeRGB /usr/share/cups/data/testprint > a.urf
```
"Gray" and "AdobeRGB" are high color depth (16 bit) and RGB is normal color depth (8 bit). All display correctly in the rasterview from above and show the correct metadata, including bit depth and color space.
[...]

PR #5226, opened 2018-01-16
When writing CUPS/PWG/Apple Raster compressed, byte-swapping never happens

tillkamppeter:
This is the fix for Issue #5225.


michaelrsweet:
I don't think this is the fix we'll want to use, but will address Issue #5225 for a subsequent 2.3 beta.

PR #5227, opened 2018-01-18
Let the PPD Generator read color spaces and color depth correctly from IPP data

tillkamppeter:
The PPD generator did not support all combinations of color spaces and color depths when reading the information from the answer to the get-printer-attributes IPP request. especially if in Apple Raster a range of color depths is given for a color space, like "W8-16", "ADOBERGB24-48", ... Now support for a color space in any color depth is recognized and the highest depth available chosen for each color space.
On my DeskJet 2540 this was especially leading to the fact that only the standard RGB color space appeared in the "ColorModel" option and not the also supported Gray and AdobeRGB color spaces. These two color spaces also print perfectly on the printer with 16 bit color depth (with CUPS 2.3b2).

[HP-DeskJet_2540-ipp-attr-ippserver.txt](https://github.com/apple/cups/files/1641166/HP-DeskJet_2540-ipp-attr-ippserver.txt)

### DIFF
--- a/cups/ppd-cache.c
+++ b/cups/ppd-cache.c
@@ -3725,7 +3725,13 @@ _ppdCreateFromIPP(char   *buffer,	/* I - Filename buffer */
 	  cupsFilePrintf(fp, "*OpenUI *ColorModel/%s: PickOne\n"
 			     "*OrderDependency: 10 AnySetup *ColorModel\n", _cupsLangString(lang, _("Color Mode")));
 
-        cupsFilePrintf(fp, "*ColorModel Gray/%s: \"<</cupsColorSpace 18/cupsBitsPerColor %s/cupsColorOrder 0/cupsCompression 0>>setpagedevice\"\n", _cupsLangString(lang, _("Grayscale")), (strstr(keyword, "16") ? "16" : "8"));
+	if (strstr(keyword, "8"))
+	{
+	  cupsFilePrintf(fp, "*ColorModel Gray/%s: \"<</cupsColorSpace 18/cupsBitsPerColor 8/cupsColorOrder 0/cupsCompression 0>>setpagedevice\"\n", _cupsLangString(lang, _("Grayscale")));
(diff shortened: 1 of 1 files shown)


michaelrsweet:
We cannot accept this pull request as-is.

First, defaulting to the highest bit depth will lead to serious performance regressions. This is not something we will change our minds on - high bit depth for AdobeRGB is OK (and expected), but not for the others.

Second, sRGB is only defined for 24-bit (8-bit per component). Higher bit depth sRGB is actually a different color space (scRGB, defined by HP and Microsoft) as none of the "tweaks" employed for 16-bit and 24-bit sRGB are valid for higher bit depths. scRGB never caught on and we don't support it "natively".

Finally, we will consider supporting 24-bit AdobeRGB, but despite the wider gamut you lose a lot of color resolution so the results are often worse than for 24-bit sRGB. For that reason we don't actually allow 24-bit AdobeRGB with Apple Raster, but if a printer is reporting support for it we might bend the rules for that...

tillkamppeter:
Thanks for the info, I simply purely followed the output of the printer on get-printer-attributes and also that CUPS, cups-filters and the printer are handling the color depths correctly.
Now I have taken your answer into account, in case of the printer offering both 8 and 16 bit for a color space defaulting to 8 bit for sRGB and 16 bit for AdobeRGB, and creating two choices for Grayscale. If the printer offers only one color depth for a color space, this depth is simply taken.
Important is also that all Black, Grayscale, sRGB, and AdobeRGB color spaces reported by the printer are recognized and especially entries in the URF info string which have two color depths.
I am ignoring DeviceGray and DeviceRGB color spaces as I do not know about how they behave color-management-wise.

michaelrsweet:
[master c283f46de] Add support for deep grayscale (ColorModel=Gray16) and 8-bit AdobeRGB (Issue #5227)

tillkamppeter:
I have tested your change now and on my printer at least it provides the same options as my change. Thanks.

Issue #5229, opened 2018-01-18
persistently substitute invalid job attributes with default values - not only in add_job


vliaskov:
Commit 6d14c1b now allows the add_job to be successfully submitted for jobs with invalid (non-utf8) job-name or requesting-user-name, by susbstituting the invalid strings with default ones ("Untitled" for job-name, "anonymous" for requesting-user-name). 

However the send_document job is still aborted during validate_user() - we get:

Returning HTTP Unauthorized for Send-Document

It looks like send_document() - >validate_user() -> cupsdIsAuthorized() is picking up the original "bad" non-utf8 requesting-user-name, although this was replaced with "anonymous" in add_job with upstream fix 6d14c1b. So I assume subsequent jobs after add_job still get the original requesting-user-name, correct?

If validate_user is commented out in send_document(), or if cupsdIsAuthorized() does the same check/replacement of requesting-user-name as commit 6d14c1b does in add_job, the problem is solved. However the first is a hack and the second code duplication.

What is the preferred way to fix the requesting user name for all jobs? Is code duplication (substituting requesting-user-name with "anonymous" in cupsdIsAuthorized() enough? Or should the attribute checks be in a separate function that will somehow be used by jobs that need it?

michaelrsweet:
@vliaskov The preferred way to fix this is to pass a valid username in the first place. Personally I would have preferred to simply reject any print request containing an invalid username, as the current workaround is to treat the request as anonymous unless "strict conformance" is enabled on the server.

michaelrsweet:
[master 092db1e29] Additional changes for the scheduler to substitute default values for invalid job attributes when running in "relaxed conformance" mode (Issue #5229)

vliaskov:
Thank you @michaelrsweet it works for me as well with this commit.

michaelrsweet:
Reopening to bring this into 2.2.x as well (since it appears the changes were not backported... :/)

michaelrsweet:
[branch-2.2 aefefc353] Backport changes for attribute substitution (Issue #5229)

PR #5234, opened 2018-01-23
cupsdIsAuthorized: substitute invalid requesting-user-name with defau…

vliaskov:
…lt value

when running in "relaxed conformance" mode (Issue #5229)

This is a potential fix for #5229. Other suggestions are welcome.


michaelrsweet:
Closing since parent bug has been otherwise fixed.

PR #5252, opened 2018-02-20
Propose fix typo

jeis2497052:
is this update OK?


michaelrsweet:
Yes, although I'll update the source and not just the generated file. Thanks for the report!

michaelrsweet:
[master 699fe0207] Update documentation (Issue #5252)

PR #5254, opened 2018-02-26
Allow raw printing using 'file' backend

michaelweghorn:
This makes it possible to do raw printing using the 'file'
backend, which was no longer working since commit
462b6e746cc2118e5e664fa95b5ee66cc702dc87.

While I am aware that this use case is not officially supported, it's still helpful to do troubleshooting in some cases without having to install any custom backend.


michaelrsweet:
Sorry, not going to do this as allowing print admins to copy files to an arbitrary location as root is not particularly safe, and we have plans of eventually removing support for file URIs completely.

michaelweghorn:
Thanks for the quick reply. Removing it completely sounds like a good option then in the long run.

PR #5287, opened 2018-04-03
Add option to have per-printer usernames in lpoptions.

Kohlrabi:
This PR implements the option to allow per-printer usernames, set via lpoptions. I avoided breaking the spec for entries in ``lpoptions``, which are of the form ``Dest <name> [<options>]`` or ``Default <name> [<options>]``. I instead added a "special" option named ``User``, which will be used by the ``options.c`` to set the cups_user, and by ``lpoptions.c`` to put the username to.

I would have loved to create a more "clean" solution, witout (ab)using options for that, but that would have probably broken the aforementioned spec of ``lpoptions``. I'm interested in better ideas of how to handle this.

### DIFF
--- a/cups/options.c
+++ b/cups/options.c
@@ -79,10 +79,7 @@ cupsAddOption(const char    *name,	/* I  - Name of option */
   else if (!_cups_strcasecmp(name, "print-quality"))
     num_options = cupsRemoveOption("cupsPrintQuality", num_options, options);
   else if (!_cups_strcasecmp(name, "User"))
-  {
     cupsSetUser(value);
-    num_options = cupsRemoveOption("User", num_options, options);
-  }
 
  /*
   * Look for an existing option with the same name...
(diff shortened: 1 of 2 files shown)


michaelrsweet:
I'm not sure we want to do this...

Kohlrabi:
I have noticed that the man-pages for lpoptions mentions an option ``-U username``, which is non-existent. I first sent a PR which removed the option from the documentation (it's currently the tip of my master). Then I thought it might be more useful to actually implement that option.

I realize that "blocking" or "overloading" the option will not allow to have a separate option "User" defined for the printer, and it is pretty non-discoverable. 

I'm willing to explore other ways to do this, and dive more deeply into the code, but I wanted to have a solution which creates 100% backwards compatible lpoptions files (unless the option "User" can be understood). And it adds new functionality.

Kohlrabi:
Another option would be to scrap the ``-U`` option altogether, and add a regular -o option to set the user name (``-o user`` or ``-o username``). In that case the parsing code in ``options.c`` can be kept, the code for ``lpoptions.c`` can be scrapped, and the dcoumentation for ``lpoptins`` and friends updated with the new option ``-o user=username``.

PR #5306, opened 2018-04-30
Fix typo in example (cupsCommand -> cupsCommands)

speachy:
Hey, this trivial patch fixes a typo in the CUPS Command File example.  I wasted several hours on this before realizing the example was wrong.


michaelrsweet:
Whoops!

speachy:
The docs also say "The only required command is PrintSelfTestPage."  Which is inconsistent with the text describing the purpose/use of 'cupsCommands'' -- and the CUPS code that utilizes it.

Cheers!

PR #5312, opened 2018-05-08
mailto: Wait for server opening message

hansmi:
A subset of all mailserver configurations reject SMTP sessions where the
client sends anything before the server sent its opening message. This
is the case with Exim on Ubuntu Artful, for example:

>   SMTP protocol synchronization error (input sent without waiting for
>   greeting)

Not waiting for the opening message is a violation of the session
initiation described in RFC5321 (section 3.1). Update the mailto
notifier to wait for a non-error status code from the server before
proceeding to start the mail transaction.


michaelrsweet:
OK, so we'll definitely pick this up, but I'm thinking I want a wait timeout in there in case we are talking to a non-conforming SMTP server...

PR #5316, opened 2018-05-22
usb-quirks: use unidir mode for for CUSTOM VK80 & VKP80III

g0hl1n:


michaelrsweet:
How common are these printers? If you just need these for a specific driver, it would be better to just bundle a quirks file with the driver - the USB backend loads everything from the /usr/share/cups/usb directory.

g0hl1n:
Hi Michael,
those printers are used a lot within POS, vending and similar devices.
But it seems to be a cleaner solution to bundle a separate quirks file with their driver.
thank you!

PR #5326, opened 2018-06-08
adding lpr --help option

zdohnal:
Hi Mike,

currently I'm doing another cleanup in CUPS on Fedora and I found some patches from which can help CUPS itself. This one is about adding --help parameter to lpr command.

Do you find it usable?

### DIFF
--- a/berkeley/lpr.c
+++ b/berkeley/lpr.c
@@ -14,6 +14,31 @@
 #include <cups/cups-private.h>
 
 
+static void
+usage (const char *name)
+{
+  _cupsLangPrintf(stdout,
+"Usage: %s [OPTION] [ file(s) ]\n"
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Maybe for 2.3.x (although the text will likely change to normalize the format with any other program help we have...)

zdohnal:
Ok, I'll look into other helps for lp* commands and try to normalize it with them as example.

michaelrsweet:
One of the things we do in other commands is to write out the options individually so that we have a few, common and short strings to localize rather than a bunch of large ones that are unique to every program.

zdohnal:

michaelrsweet:
OK, I've pushed a full set of changes so that all user commands now support --help.

PR #5328, opened 2018-06-08
adds logging when job fails due bad required conversion

zdohnal:
This one adds messages to the log when job failed due bad required conversion and proposes available conversions for usage.

### DIFF
--- a/scheduler/job.c
+++ b/scheduler/job.c
@@ -610,9 +610,27 @@ cupsdContinueJob(cupsd_job_t *job)	/* I - Job */
 
     if (!filters)
     {
+      mime_filter_t *current;
+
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Well, we should never get in this state since we won’t accept a job in a format we can’t convert... will have to think on this one...

Also, we already log the filters on startup when debug logging is enabled. Don’t think we want to log them again for every failed job (a lot of log entries)

zdohnal:
Sounds reasonable... then what about removing the first message and have the latter for DEBUG2 level?

zdohnal:

zdohnal:
Ok, I re-evaluated  the issue and it doesn't seem worth the pros.

PR #5329, opened 2018-06-08
Add device ID for Dymo LabelWriter 400

zdohnal:
SSIA

### DIFF
--- a/ppdc/sample.drv
+++ b/ppdc/sample.drv
@@ -128,6 +128,7 @@ Version "2.3"
 {
   Manufacturer "Dymo"
   ModelName "Label Printer"
+  Attribute "1284DeviceID" "" "MFG:DYMO;MDL:LabelWriter 400;"
   Attribute NickName "" "Dymo Label Printer"
   PCFileName "dymo.ppd"
   DriverType label
--- a/ppdc/sample.drv
+++ b/ppdc/sample.drv
@@ -128,7 +128,6 @@ Version "2.3"
 {
   Manufacturer "Dymo"
   ModelName "Label Printer"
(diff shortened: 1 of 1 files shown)


michaelrsweet:
There are other printers this driver works with, so I'm not sure I want to tie this driver to it...

Is there a reason you need this?

zdohnal:
I'm sorry, you're right - I made a mistake - I reviewed the patch badly, setting the ppd for only LabelWriter is not okay...
About the reason: Once there was Fedora user, who had a problem with installing his Dymo LabelWriter 400 - he got a bad ppd file and Tim fixed it by the patch above. There is the report https://bugzilla.redhat.com/show_bug.cgi?id=747866 . Dymo LabelWriter 400 doesn't have any support in foomatic or gutenprint and cups dymo ppd worked for user.
Is it acceptable to have another ppd in CUPS for LabelWriter (see updated patch) or should I pursue this issue in f.e. foomatic, because ppds are going away?

zdohnal:
@michaelrsweet ok, I understood the printer driver projects will stay, because printer applications (ipp server + printer driver project such Gutenprint/Foomatic) can have use for them when managing old legacy printers, which don't support IPP everywhere...
Is there a way in .drv file to tell: "Please use this ppd with this printer too"? I was unable to find such way... if there is not, we can close the pull request.

michaelrsweet:
@zdohnal Printer driver projects, yes, but I'm (perhaps too optimistically) hoping that Foomatic will no longer be needed.

The 1284DeviceID and NickName strings are the ones used for driver matching. We aren't technically adding another driver here, but historically we haven't provided printer-specific values. Maybe we can do something with the MFG and CMD values so we match any Dymo printer?

zdohnal:

PR #5337, opened 2018-06-11
log history for journal

zdohnal:
Hi Mike,
when I was reproducing Fedora bug https://bugzilla.redhat.com/show_bug.cgi?id=1589593 , I found out the LogLevel directive is ignored when I have logging set to syslog (systemd-journald in Fedora) when LogDebugHistory is greater than 0 (I have it set to 200, according gdb).
To track the fix I looked more into cupsdLogJob function and found out there is job logs history tracked for no-systemd OS - so to solve 'ignoring issue' I implemented this patch, which now correctly respects the LogLevels (tested) and has job logs history for systemd OS (this one needs testing).
I tested it by changing where ErrorLog goes (syslog, /var/log/cups) and changing LogLevels (none, warn, debug2), but I don't know how to test job logs history feature.
Would CUPS benefit from this patch?

### DIFF
--- a/scheduler/log.c
+++ b/scheduler/log.c
@@ -565,48 +565,17 @@ cupsdLogJob(cupsd_job_t *job,		/* I - Job */
     return (1);
 
 #ifdef HAVE_SYSTEMD_SD_JOURNAL_H
-  if (!strcmp(ErrorLog, "syslog"))
-  {
-    cupsd_printer_t *printer = job ? (job->printer ? job->printer : (job->dest ? cupsdFindDest(job->dest) : NULL)) : NULL;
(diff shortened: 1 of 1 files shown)


michaelrsweet:
I'll need to look at this a little more, but I *think* we'll take this.

michaelrsweet:
Pushed a slightly different change:

[master 945c83b90] Fix auto-debug logging of job errors with systemd (Issue #5337)

[branch-2.2 ae9f0b239] Fix auto-debug logging of job errors with systemd (Issue #5337)

zdohnal:
LGTM! Thanks, Mike!

Issue #5315, opened 2018-05-16
fix or document proper configuration for specifying default output order


archenemies:
I have an HP Photosmart Plus B210a which prints pages face up. I am trying to figure out how to configure Cups to print documents so that I don't have to reverse them by hand when they come out of the printer.

I found [a discussion on linuxquestions.org](https://www.linuxquestions.org/questions/ubuntu-63/can-i-set-the-cups-output-order-to-always-print-in-reverse-order-746843/) which summarizes a recommended fix:

     Add the line
     *DefaultOutputOrder: "reverse"
     to the file
     /etc/cups/ppd/PrinterName.ppd

I have tried this with varying results. I printed twelve two-page documents to try to investigate what is happening. I tried printing via the `lp` command, via Evince and via Okular. I tried using a ppd file with and without the `DefaultOutputOrder` line, and I also tried specifying a default output order with `lpoptions` of "reverse" or "normal":

    lpoptions -dHP_Photosmart_Plus_B210a -o outputorder=reverse
    ...
    lpoptions -dHP_Photosmart_Plus_B210a -o outputorder=normal

The results:

* `lp`: Ignores ppd line, honors lpoptions setting
* Evince: Ignores lpoptions setting, honors ppd
* Okular: Ignores both lpoptions and ppd

As with the reporter of #1679, I would have expected the PPD which is downloaded by default for my printer to give the "correct" behavior by default, which is to say, not requiring me to manually reverse long documents when they come out of the printer.

I also tried the solution of #1679, using

    lpadmin -p printer -o outputorder-default=reverse
    lpadmin -p printer -o outputorder-default=normal

This setting was honored by Evince but not by `lp` or Okular.

Okular's print dialog allows the user to change the ouput ordering, but Evince's print dialog, which appears more "standard" to me, says "Page Ordering: Not available".

There was some mention of all documents going through `pstops` as a final filter. If this is true then it should be possible for Cups to give users a single point at which to configure output ordering, which works uniformly across all applications which might submit print jobs to Cups. In either case it should be documented what is the preferred way to deal with face-up printers.

[...]

PR #5345, opened 2018-06-29
PPDs for driverless printing: Make jobs be printed in correct order for face-up/face-down

tillkamppeter:
In https://github.com/apple/cups/issues/5315 and https://github.com/OpenPrinting/cups-filters/issues/47 a user complained about that on printers which stack up their output pages face-up the pages should get printed in reverse order.
I have investigated this and the main problem is that many PPD files are missing the appropriate information about how the paper gets stacked.
In case of driverless printing the answers to the get-printer-attributes request always contains the needed information. Either the names of the output-bins (attributes "output-bin-supported", "output-bin-default" are named "face-up" or "face-down" or there is an array named "printer-output-tray" with the same number of items as "output-bin-supported", describing the properties of each output bin, especially telling about the way the pages get stacked ([PWG 5100.13 – IPP: Job and Printer Extensions – Set 3](https://ftp.pwg.org/pub/pwg/candidates/cs-ippjobprinterext3v10-20120727-5100.13.pdf), section 5.6.36).
The patch (which I also have applied to the PPD generator of cups-filters) makes this information available in the generated PPD files for driverless printing. Once, the PPD file gets a "*DefaultOutputOrder: ..." line with either "Normal" or "Reverse" as value, the latter indicating that the pages get stackked face-up in the default tray and therefore the pages need to get printed in reverse order (which `pdftopdf` of cups-filters takes care of in this case). Second, if the PPD has an "OutputBin" option, I add a "*PageStackOrder ..." line to each choice of the "OutputBin" option telling whether this bin is face-up or face-down. Also this is taken care of by `pdftopdf` (see "*OutputOrder", p124, and "*PageStackOrder", p125, in Adobe's PPD specs).
This way printers set up with driverless printing of CUPS should print the jobs in the correct order with respect to the fact that they stack their output face-up or face-down.

### DIFF
--- a/cups/ppd-cache.c
+++ b/cups/ppd-cache.c
@@ -3034,6 +3034,7 @@ _ppdCreateFromIPP(char   *buffer,	/* I - Filename buffer */
   cups_array_t		*sizes;		/* Media sizes supported by printer */
   cups_size_t		*size;		/* Current media size */
   ipp_attribute_t	*attr,		/* xxx-supported */
+			*attr2,
 			*defattr,	/* xxx-default */
                         *quality,	/* print-quality-supported */
 			*x_dim, *y_dim;	/* Media dimensions */
@@ -3071,7 +3072,12 @@ _ppdCreateFromIPP(char   *buffer,	/* I - Filename buffer */
 					/* Locale data */
   cups_array_t		*fin_options = NULL;
 					/* Finishing options */
-
+  char			*defaultoutbin = NULL;
+  const char		*outbin,
+			*outbin_properties;
+  int			outputorderinfofound = 0,
+			faceupdown = 1,
+			firsttolast = 1;
 
  /*
   * Range check input...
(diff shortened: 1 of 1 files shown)


cups/ppd-cache.c
```diff
@@ -3138,6 +3144,51 @@ _ppdCreateFromIPP(char   *buffer,	/* I - Filename buffer */
   cupsFilePrintf(fp, "*NickName: \"%s - IPP Everywhere\"\n", model);
   cupsFilePrintf(fp, "*ShortNickName: \"%s - IPP Everywhere\"\n", model);
 
+  /* Which is the default output bin? */
+  if ((attr = ippFindAttribute(response, "output-bin-default", IPP_TAG_MIMETYPE)) != NULL)
+    defaultoutbin = strdup(ippGetString(attr, 0, NULL));
+  /* Find out on which position of the list of output bins the default one is, if there
+     is no default bin, take the first of this list */
+  i = 0;
+  if ((attr = ippFindAttribute(response, "output-bin-supported", IPP_TAG_MIMETYPE)) != NULL)
```
michaelrsweet: This should be IPP_TAG_ZERO to allow for IPP_TAG_KEYWORD or IPP_TAG_NAME values. But not IPP_TAG_MIMETYPE (which is the mimeMediaType syntax...)

cups/ppd-cache.c
```diff
@@ -3138,6 +3144,51 @@ _ppdCreateFromIPP(char   *buffer,	/* I - Filename buffer */
...
+      {
+	defaultoutbin = strdup(outbin);
+	break;
+      }
+      else if (strcasecmp(outbin, defaultoutbin) == 0)
+	break;
+    }
+  }
+  /* Look up the corresponding entry in the "printer-output-tray" attribute */
+  if ((attr = ippFindAttribute(response, "printer-output-tray", IPP_TAG_MIMETYPE)) != NULL &&
```
michaelrsweet: printer-output-tray is of type 1setOf octetString, so you want IPP_TAG_STRING here. And you'll need to use ippGetOctetString, copy it to a C string, and then look for the properties.

michaelrsweet:
Till, I appreciate the pull request but the current changes have issues. See my comments...

tillkamppeter:
I have fixed the two issues now.

tudor2k3:
Hello,

     Is this also a fix for my Canon IR C1225 not respecting long-edge/short-edge duplex printing? Or do i have to open a different bug?

                     Thank you.

tillkamppeter:
@tudor2k3, please open a separate issue, as this one is only for OutputOrder as it does not do more than adding lines concerning the output order to the PPD.

tillkamppeter:
Is there still anything to be fixed on my pull request?

michaelrsweet:
Till, I haven't had a chance to review this fully, and since I am on a summer break at the moment it won't be happening anytime soon. Don't expect anything to happen until the end of August at the earliest.

michaelrsweet:
@tillkamppeter You are still using the non-portable strcasestr function...  Since these values are case sensitive, it should just be a matter of changing to strstr.

There are some other potential optimizations I'll look at doing as well...

michaelrsweet:
[master bf86060f9] Support face-up printers (Issue #5345)

[branch-2.2 fe1fac144] Support face-up printers (Issue #5345)

PR #5369, opened 2018-08-07
Update man-lpadmin.html to document -E as option

JackGrinningCat:
Added -E before -x -p -d as documented option.
Added example with description of usage of -E as enabled and -E for enforced encryption.


michaelrsweet:
I didn't end up adding an example, just documentation for the option and a note about the double meaning.

[master d5d647780] Update lpadmin man page (Issue #5369)

[branch-2.2 b42427212] Update lpadmin man page (Issue #5369)

PR #5374, opened 2018-08-15
Fixes based on coverity scan issues

zdohnal:
Hi Mike,
I ran static analysis by coverity scan for cups-2.2.6, which we have in Fedora 28, and fixed several issues which were reported. I'll send you initial results of scan by email (in case there are some issues, which can be exploitable).
Does it look good to you?


zdohnal:
I'm only sending fixes for issues which are in master branch too.

zdohnal:
I'll send cleaner branch...

zdohnal:
See #5375

PR #5375, opened 2018-08-15
Fixing several issues reported by coverity scan

zdohnal:
Here is cleaner commit...

### DIFF
--- a/backend/ipp.c
+++ b/backend/ipp.c
@@ -3624,4 +3624,7 @@ update_reasons(ipp_attribute_t *attr,	/* I - printer-state-reasons or NULL */
     fprintf(stderr, "%s\n", add);
   else if (rem[0])
     fprintf(stderr, "%s\n", rem);
+
+  if (new_reasons != NULL)
+    cupsArrayDelete(new_reasons);
 }
--- a/cups/http.c
+++ b/cups/http.c
@@ -3916,6 +3916,7 @@ http_create(
   {
     _cupsSetError(IPP_STATUS_ERROR_INTERNAL, strerror(errno), 0);
     httpAddrFreeList(addrlist);
+    httpAddrFreeList(myaddrlist);
     return (NULL);
   }
 
--- a/cups/raster.c
+++ b/cups/raster.c
@@ -1076,7 +1076,7 @@ cupsRasterWriteHeader(
 
     void *dst = fh.cupsReal; /* Bypass bogus compiler warning */
     void *src = r->header.cupsReal;
-    memcpy(dst, src, sizeof(fh.cupsReal) + sizeof(fh.cupsString));
+    memcpy(dst, src, sizeof(fh.cupsReal));
 					/* VendorData */
 
     strlcpy(fh.cupsRenderingIntent, r->header.cupsRenderingIntent,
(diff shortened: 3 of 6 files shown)


zdohnal:
All issues in report which sent you were labeled as important, so there are other issues, but coverity doesn't count them as important.

backend/ipp.c
```diff
@@ -3624,4 +3624,7 @@ update_reasons(ipp_attribute_t *attr,	/* I - printer-state-reasons or NULL */
     fprintf(stderr, "%s\n", add);
   else if (rem[0])
     fprintf(stderr, "%s\n", rem);
+
+  if (new_reasons != NULL)
+    cupsArrayDelete(new_reasons);
```
michaelrsweet: OK, we should just initialize new_reasons to NULL up top and then call cupsArrayDelete unconditionally (it handles NULLs just fine.)

cups/http.c
```diff
@@ -3916,6 +3916,7 @@ http_create(
   {
     _cupsSetError(IPP_STATUS_ERROR_INTERNAL, strerror(errno), 0);
     httpAddrFreeList(addrlist);
+    httpAddrFreeList(myaddrlist);
```
michaelrsweet: I need to look at this more closely; my gut says we should just be freeing myaddlist and not addrlist (since, if supplied, we copy addrlist to myaddrlist...)

cups/raster.c
```diff
@@ -1076,7 +1076,7 @@ cupsRasterWriteHeader(
 
     void *dst = fh.cupsReal; /* Bypass bogus compiler warning */
     void *src = r->header.cupsReal;
-    memcpy(dst, src, sizeof(fh.cupsReal) + sizeof(fh.cupsString));
+    memcpy(dst, src, sizeof(fh.cupsReal));
```
michaelrsweet: This is wrong. We need to copy everything in cupsReal and cupsString - cupsString follows cupsReal in the structure so this is valid.

scheduler/cups-driverd.cxx
```diff
@@ -1555,6 +1555,24 @@ list_ppds(int        request_id,	/* I - Request ID */
...
+    cupsArrayDelete(matches);
+  if (requested != NULL)
+    cupsArrayDelete(requested);
+
+  /*
+   * Free device_id_re and make_and_model_re, no longer needed
+   */
+  free(device_id_re);
+  free(make_and_model_re);
+
```
michaelrsweet: We don't care about this because main() calls list_ppds() and then returns, freeing all memory used by the program. Perhaps I should have it not return and instead exit().

michaelrsweet:
I pushed changes for the full scan report to branch-2.2 and master:

[master 3416fe952] Fix memory leaks found by Coverity (Issue #5375)

[branch-2.2 1a3ff20f7] Fix memory leaks found by Coverity (Issue #5375)

zdohnal:
I apologize for late reply, I was on vacation last week - I'll run it through coverity scan and I'll check with you about results.

zdohnal:
I checked the commit with coverity - the results are okay, no new important reports. Thank you!

PR #5393, opened 2018-09-04
cc->ssl_options needs to be initialized

zdohnal:
Hi Mike,
I was backporting MinTLS/MaxTLS functionality from 2.2.8 to 2.2.6 and I found out when you don't have any 'SSLOptions' directive in client.conf, then cc->ssl_options isn't initialized and has undefined value. And that undefined value is used in other functions and can cause troubles.
Would you mind merging this patch? It is just initialization and I and user who had troubles tested it and it worked.


michaelrsweet:
Um, that function does a memset of the whole conf struct to 0. Initializing ssl_options to 0 here is unnecessary...

PR #5398, opened 2018-09-17
Create more general message from job.c

zdohnal:
Hi Mike,

some users ( https://bugzilla.redhat.com/show_bug.cgi?id=1629677 ) can be confused by messages from job.c about seeing logs in error_log, but system default is set to syslog. I tried to create more general message with mentioning of cups-files.conf:

"XXXXXXXXXXXXXXXXXXXXX; please consult logs for details. See where log location is in cups-files.conf."

Is it acceptable for CUPS to accept such change in the message? I hope it can help against user's confusion...

### DIFF
--- a/scheduler/job.c
+++ b/scheduler/job.c
@@ -3337,7 +3337,7 @@ finalize_job(cupsd_job_t *job,		/* I - Job */
 	  {
 	    job_state = IPP_JOB_ABORTED;
 	    message   = "Job aborted due to backend errors; please consult "
-	                "the error_log file for details.";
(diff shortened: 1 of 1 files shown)


michaelrsweet:
No reason we can't just report the location...

michaelrsweet:
[branch-2.2 b485551a7] Log actual location of error_log (Issue #5398)

zdohnal:
You're faster than me :)  - but I'm glad I would solved it the same way as you after you give an idea about telling actual place of logs. But I forgot about emptiness of ErrorLog, so still way to go.

PR #5408, opened 2018-10-04
Update index.html.in

victor-oliveira1:
The translation is wrong.
Instead of Administrarção it's Administração.


michaelrsweet:
[master 08eac601d] Localization updates (Issue #5408)

[branch-2.2 37665f15b] Localization updates (Issue #5408)

PR #5411, opened 2018-10-13
Update build files

dannftk:
Updated configure, several config-scripts/*.m4, several Makefiles

- in config-scripts/cups-common.m4 'AR' variable is used from environment (required for object-files compiled with -flto), in case it is not provided AR is set by searching for 'ar' program (by default as it was)
- in config-scripts/cups-compiler.m4 'Os' imposed flag has been removed from OPTIM variable because it interferes with '-flto' if provided. 'Os', if required, can be improsed by CFLAGS/CXXFLAGS environment variables configurable by a user
- in config-scripts/cups-opsys.m4 'CODE_SIGN' variable is set by searching for PATH to 'true' binary instead of using forcefully imposed '/usr/bin/true' because in some systems 'true' can be located in a different place. If 'true' cannot be found by this way the default behaviour is preserved and takes place (like it was)
- in cups/Makefile, ppdc/Makefile, filter/Makefile, scheduler/Makefile, cgi-bin/Makefile 'LDFLAGS' is used when linking a shared library for being able to link at least with '-flto' flag. LDFLAGS should be provided always when linking, LDFLAGS as well as CFLAGS and CXXFLAGS can be set by a user whom configures a project
- 'configure' script updated with 'autoconf --force > configure', version 2.69 of autoconf is used


dannftk:
I checked building by means of the following command
```
AR=llvm-ar CC=clang CXX=clang++ CFLAGS="-O2 -pipe -march=native -mtune=generic -flto -std=gnu11" CXXFLAGS="-O2 -pipe -march=native -mtune=generic -flto -std=gnu++17 -stdlib=libc++" LDFLAGS="-flto -fuse-ld=gold" ./configure
```
then I executed
```
make all
```
everything has been built successfully

Actually, I didn't touch any *.dylib and libtool specific targets in Makefile, for me to haven't been able to verify them properly in my environment, let it alone I don't have Mac to check at least *.dylib, let it be the next iteration

I could check libtool specific targets and improve them if you like me to do that. For this I should adjust my environment

michaelrsweet:
Can’t accept this as-is, but will review in detail next week.

dannftk:
I will be waiting for your review up impatiently

config-scripts/cups-common.m4
```diff
@@ -34,13 +34,18 @@ CPPFLAGS="${CPPFLAGS:=}"
 CXXFLAGS="${CXXFLAGS:=}"
 LDFLAGS="${LDFLAGS:=}"
 
+dnl Setting compiler environment
+AR="${AR:=}"
+
```
michaelrsweet: Autoconf's generated configure script already allows the environment variable to override the configure check.

config-scripts/cups-compiler.m4
```diff
@@ -98,7 +98,7 @@ if test -n "$GCC"; then
 	if test -z "$OPTIM"; then
 		if test "x$with_optim" = x; then
 			# Default to optimize-for-size and debug
-       			OPTIM="-Os -g"
+       			OPTIM="-g"
```
michaelrsweet: The default compiler options need to specify optimization. Specify --with-optim="-g" to force a debug-only build, or to pass any other optimization options.

config-scripts/cups-opsys.m4
```diff
@@ -37,9 +37,10 @@ fi
 AC_SUBST(LOCALTARGET)
 
 AC_MSG_CHECKING(for codesign utility)
-CODE_SIGN="/usr/bin/true"
-AC_SUBST(CODE_SIGN)
 if test "$host_os_name" = darwin; then
 	CODE_SIGN="/usr/bin/codesign"
+else
```
michaelrsweet: I think I'll use:

    AC_PATH_PROGS(CODE_SIGN, codesign true)

dannftk: what if none is found? I think we should explicitly fail the build

michaelrsweet:
[master c8c5ec3b5] Search for codesign/true, use LDFLAGS for shared libraries (Issue #5411)

config-scripts/cups-common.m4
dannftk: I tried to pass these environment variables in 'configure' script without my changes and 'ar' is obtained from the system by searching for 'ar' program using PATH:
```
AR=llvm-ar CC=clang CXX=clang++ CFLAGS=-flto CXXFLAGS=-flto LDFLAGS=-flto ./configure
```
at the end of the build I see
```
*/usr/bin/ar* crvs libcups.a adminutil.o array.o auth.o backchannel.o backend.o debug.o dest.o dest-job.o dest-localization.o dest-options.o dir.o encode.o file.o getdevices.o getifaddrs.o getput
file.o globals.o hash.o http.o http-addr.o http-addrlist.o http-support.o ipp.o ipp-support.o langprintf.o language.o md5.o md5passwd.o notify.o options.o ppd.o ppd-attr.o ppd-cache.o ppd-conflicts.o ppd-custom.o ppd-emit.o ppd-localize.o ppd-mark.o ppd-page.o ppd-util.o pwg-media.o request.o sidechannel.o snmp.o snprintf.o string.o tempfile.o thread.o tls.o transcode.o usersys.o util.o
```

[here](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Programs.html) manual says, quote, 
> Macro: AC_PATH_PROGS (variable, progs-to-check-for, [value-if-not-found], [path = ‘$PATH’])
> > Like AC_CHECK_PROGS, but if any of progs-to-check-for are found, set variable to the absolute name of the program found. *The result of this test can be overridden by setting the _variable_ variable*. A positive result of this test is cached in the ac_cv_path_variable variable.

I think it does not mean that the result can be ignored in a favor of an environment variable with the same name, does it?

config-scripts/cups-compiler.m4
dannftk: Ok, maybe, I would understand if you specify -O2 by default, ok, it is understandable, it is also by convention, like a trade-off in optimization levels. But -Os looks weird. No offence given

dannftk: take a look how [it is done in cmake](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html)
it can be food for mind

Issue #5359, opened 2018-07-27
Jobs with multiple files don't complete when backend fails


bryan-mason:
When the following conditions are met:

- A print job is submitted with multiple files (i.e. lp -d `pq /etc/services /etc/services`)
- The print job uses at least one filter (i.e. not a raw queue)
- The backend fails (due to, for example, a Broken Pipe error, which is a common occurrence)

Then the job stays in the queue.  The backend fails, but the filter doesn't receive SIGPIPE and never exits, and the scheduler doesn't kills the job.

If the print job has only one file, then the filter gets SIGPIPE when the backend fails and the job is aborted (or retried depending on the Error Policy).

I've reproduced this in CUPS 1.4.2, 1.6.3 and 2.2.6.

I've been trying (unsuccessfully) to discover a difference in the way the scheduler starts filters for multi-file jobs vs. single-file jobs that would account for this behavior.  Suggestions for where to look in the code would be appreciated (as would a patch to fix this, but I'm happy to work on the patch myself if I can get a push in the right direction).

Or would it make sense to call `stop_job()` somewhere if the backend fails.  There's not much point in continuing to process a job if the backend has failed, is there?

Thanks.

michaelrsweet:
@bryan-mason "Broken pipe" should not be a common occurrence for a backend, particularly when all of the standard backends block/handle it.

The difference with single-vs-multi document jobs is that the backend (and associated pipes) remains active for all of the documents in the job - basically we reuse them for all documents, e.g:

    filters for document 1  \
    filters for document 2   | backend
    ...                                    /
    filters for document N /

Anyways, we should probably be closing the pipes when the backend fails, which will allow the filters to see that the backend has gone away (rather than block on IO) and allow the job to abort.

bryan-mason:
>  "Broken pipe" should not be a common occurrence for a backend, particularly when all of the standard backends block/handle it.

It's been my experience supporting enterprise customers that:

    D [25/Jul/2018:17:24:05 -0700] [Job 42] Error reading back-channel data: Connection reset by peer
    E [25/Jul/2018:17:24:05 -0700] [Job 42] Unable to write print data: Broken pipe

Is one of the more common failure modes that people report.  The cause always seems to be some sort of network equipment problem, and the socket backend handles it gracefully and exits cleanly, but it's still a problem (usually because `ErrorPolicy` is `stop-printer` and the customer wants to know why their print queue stopped).

michaelrsweet:
[...]

PR #5413, opened 2018-10-17
Abort job with multiple files when backend fails

zdohnal:
Hi,

there is the first shot, which gets multiple files job aborted when backend suddenly lost connection during data transfer as I wrote in email. I am lost in making such job retry - I think a recovery for retry is rather complicated - checking which file didn't print correctly and probably creating new job for the unfinished business.
This pull request is connected to issue #5359 , is it a right way to go?

### DIFF
--- a/backend/socket.c
+++ b/backend/socket.c
@@ -443,7 +443,7 @@ main(int  argc,				/* I - Number of command-line arguments (6 or 7) */
   if (print_fd != 0)
     close(print_fd);
 
-  return (CUPS_BACKEND_OK);
+  return (tbytes < 0 ? CUPS_BACKEND_FAILED : CUPS_BACKEND_OK);
 }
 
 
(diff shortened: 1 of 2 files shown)


zdohnal:
I ran basic tests from project and results are good. There are expected failures (according test) and one unexpected, but without connection to the issue (4.4-subscription-ops.test failed with server-error-internal-error (Invalid group tag.))

michaelrsweet:
I pushed a slightly different fix, and all tests appear to pass.

[master 72a213482] Fix stuck multi-file jobs (Issue #5359, Issue #5413)

[branch-2.2 e7e33bf64] Fix stuck multi-file jobs (Issue #5359, Issue #5413)

zdohnal:
Thank you, Mike! I'll test it tomorrow!

PR #5435, opened 2018-11-12
Avoid truncation when encoding Kerberos token

hansmi:
The "httpEncode64_2" function appends padding (0-3x "="). The buffer
size calculation in "_cupsSetNegotiateAuthString" did the calculation
wrongly and would have a buffer overflow for tokens of size (N * 4)
\+ 1 and (N * 4) + 2. With this change the buffer size is computed
correctly.

See commit message for detailed calculation.

### DIFF
--- a/cups/auth.c
+++ b/cups/auth.c
@@ -442,7 +442,7 @@ _cupsSetNegotiateAuthString(
     */
 
     int authsize = 10 +			/* "Negotiate " */
-		   (int)output_token.length * 4 / 3 + 1 + 1;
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Changing the title to reflect the issue - httpEncode64_2 won't overflow, it will truncate.

michaelrsweet:
[master ef2f369c0] Fix potential truncation of Kerberos credentials (Issue #5435)

[branch-2.2 fe35d60e6] Fix potential truncation of Kerberos credentials (Issue #5435)

PR #5451, opened 2018-12-08
Fix call to usage which requires an int argument

Polynomial-C:
Otherwise compilation fails:

```
main.c: In function ‘main’:
main.c:158:7: error: too few arguments to function ‘usage’
       usage();
       ^~~~~
main.c:73:14: note: declared here
 static void  usage(int status) _CUPS_NORETURN;
              ^~~~~
make[1]: *** [../Makedefs:270: main.o] Error 1
```


michaelrsweet:
[master 065632e20] Fix compile errors - not sure why these didn't show up before...

PR #5473, opened 2019-01-08
PPD parser: Calling cupsFileGetChar(...) after EOF

ppawliczek:
For some input data the function cupsFileGetChar(...) was incorrectly called
from the function ppd_read(...) after returning EOF in a previous call. This
causes incorrect state of cups_file and results in memory leak.

### DIFF
--- a/cups/ppd.c
+++ b/cups/ppd.c
@@ -3038,7 +3038,7 @@ ppd_read(cups_file_t    *fp,		/* I - File to read from */
       }
     }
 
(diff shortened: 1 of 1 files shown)


michaelrsweet:
That's not the right place to add this check. I've committed a change to the cupsFileGetChar and cupsFileRead functions to immediately return EOF rather than calling cups_fill.

[master 96fee3038] Protect against continuing to read from a file at EOF (Issue #5473)

PR #5475, opened 2019-01-08
PPD parser: Only last "Emulators" statement is taken into account + memory leak

ppawliczek:
Each occurence of "Emulators" keyword overwrites the previous parsed
"Emulators" statement. It also causes a memory leak. The problem is solved here
by concatenating together values from all occurences of "Emulators" statements.

### DIFF
--- a/cups/ppd.c
+++ b/cups/ppd.c
@@ -470,6 +470,7 @@ _ppdOpen(
 			ll_CC[7];	/* Language w/country + '.' */
   size_t		ll_len = 0,	/* Base language length */
 			ll_CC_len = 0;	/* Language w/country length */
+  void			*temp_ptr;	/* Temporary variable with a pointer to resized buffer */
   static const char * const ui_keywords[] =
 			{
 #ifdef CUPS_USE_FULL_UI_KEYWORDS_LIST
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Happy to fix this, but keep in mind that CUPS never uses the value and PPDs are long deprecated...

cups/ppd.c
```diff
@@ -654,6 +655,7 @@ _ppdOpen(
   ppd->landscape      = -90;
   ppd->coptions       = cupsArrayNew((cups_array_func_t)ppd_compare_coptions,
                                      NULL);
+  ppd->emulations     = NULL;
```
michaelrsweet: Unnecessary because ppd_file_t structure is zeroed.

cups/ppd.c
```diff
@@ -1202,27 +1204,32 @@ _ppdOpen(
 	    sptr ++;
 	}
 
-      ppd->num_emulations = count;
-      if ((ppd->emulations = calloc((size_t)count, sizeof(ppd_emul_t))) == NULL)
+      temp_ptr = realloc(ppd->emulations, (size_t)(ppd->num_emulations+count)*sizeof(ppd_emul_t));
```
michaelrsweet: Sadly, CUPS runs on platforms where `realloc(NULL, size)` will crash hard. But I can refactor this.

michaelrsweet:
I opted to just remove the code that deals with the emulations array completely - nothing in CUPS uses this array and it was just more "dead code" cluttering up libcups.

[master 4b5a6914f] Stop parsing the Emulators keywords in PPD files (Issue #5475)

[branch-2.2 558bba72f] Stop parsing the Emulators keywords in PPD files (Issue #5475)

PR #5477, opened 2019-01-10
Raw TLS socket support

mcb30:
This small patch series adds support for TLS to the socket backend.  This allows CUPS to communicate with printers such as the Zebra ZQ520, which can listen on port 9143 for TLS-encrypted AppSocket/JetDirect connections.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ backend/socket
 backend/test1284
 backend/testbackend
 backend/testsupplies
+backend/tlsraw
 backend/usb
 berkeley/lpc
 berkeley/lpq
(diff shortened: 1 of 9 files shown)


michaelrsweet:
Sorry, we have no desire to support yet another vendor specific ‘protocol’.

mcb30:
It's just standard AppSocket/JetDirect but with TLS.  It's not vendor-specific in any way: the example of the Zebra ZQ520 is just to demonstrate that there are printers in existence that support this.

michaelrsweet:
First, this is the first printer I've ever heard of supporting raw TLS sockets - it isn't common, nor am I aware of any standardization efforts or protocol documentation for it. (And no, the TLS spec isn't enough!)

Second, there *is* an international standard protocol with TLS support that is supported by over 98% of all network printers - IPP. Besides the basic print data transfer supported by IPP, you also get status monitoring, capabilities, and (in most cases) higher-level file format support allowing printers to "just work" without vendor-specific software.

So, we have no interest in supporting a vendor-specific protocol that continues a legacy of vendor-specific software.

PR #5507, opened 2019-02-03
scheduler/main.c: support READY_FD protocol

CameronNemo:
This pull request add supports for a lightweight service readiness protocol.

The specification and reference supervisor implementation of the protocol can be found [here](https://gitlab.com/chinstrap/ready).

The protocol is compatible with the s6 service supervision suite, and the reference implementation can be paired with many other service supervisors.

### DIFF
--- a/man/cupsd.man.in
+++ b/man/cupsd.man.in
@@ -23,6 +23,8 @@ cupsd \- cups scheduler
 ] [
 .B \-l
 ] [
+.B \-r
+] [
 .B \-s
 .I cups-files.conf
 ] [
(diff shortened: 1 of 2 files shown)


michaelrsweet:
Well, at this point I'm not keen on these changes as-is, and it isn't clear at all how cupsd is started or managed using this new interface?

CameronNemo:
Is there any way I can approve the patch to be less intrusive or more palatable?

Regarding how cupsd is started using a supervisor supporting the protocol: that is really up to service manager. You can see some examples [here](https://gitlab.com/chinstrap/ready/blob/master/docs/man/ready.8.md#examples). Note that only s6 supports the mechanism natively, other service managers would need to implement the supervision side of the protocol or use the reference implementation (which itself is a small, flexible, portable and liberally licensed program written in C).

skarnet:
s6 is already pretty popular with Docker users. You can look up "s6 Docker" to get an idea.
[Badoo](https://badoo.com/) uses it. Lots of small companies use it. It's not the default in any mainstream distribution at the moment, but expect that to change in the upcoming years. ;-)

Conan-Kudo:
What makes a file descriptor better than a socket?

michaelrsweet:

PR #5513, opened 2019-02-11
Show warning about logs in syslog in Web UI

zdohnal:
Hi Mike!
I set all cupsd logs to be in system log (journal), but it breaks viewing logs in CUPS web ui - it ends with 'Not Found', which can be confusing for users.
I tried to focus on the issue and solve it in the pull request - I introduced new file in CUPS_CACHEDIR, which will contain warning message (file is created and the message is written there during reading configuration - if configuration changes in the way there is no logs in syslog, file is deleted):

'The log has been moved to system log. Please consult your logging system for logs.'

CGI binary then gets the file instead of original log file (if CUPS is built with correct features - systemd or vsyslog and the logging is set to syslog).

Viewing logs in web ui shows warning instead of 'Not Found'. Is it acceptable behavior and could it be added to the project?

### DIFF
--- a/scheduler/client.c
+++ b/scheduler/client.c
@@ -2814,7 +2814,7 @@ get_file(cupsd_client_t *con,		/* I  - Client connection */
       if (AccessLog[0] == '/')
         strlcpy(filename, AccessLog, len);
 #ifdef HAVE_SYSTEMD_SD_JOURNAL_H || defined(HAVE_VSYSLOG)
-      else if (!strcmp(AccessLog, "syslog"))
+      else if (!strcmp(AccessLog, "syslog") && WebInterface)
       {
         cupsdLogClient(con, CUPSD_LOG_DEBUG2, "Log is set to syslog, get syslog warning file \"%s\".", SyslogWarn);
         strlcpy(filename, SyslogWarn, len);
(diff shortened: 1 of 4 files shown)


michaelrsweet:
I'd much rather just hide the log buttons when syslog/etc. are in use.

zdohnal:
@michaelrsweet You're right, that's better idea, although I found two things:

1) As I looked into code and if I understood it correctly, it would need to read cups-files.conf in cupsAdminGetServerSettings() function to get current setup of log location. Or move 'AccessLog/ErrorLog/PageLog' directives to cupsd.conf/create a new directive in cupsd.conf for it.

2) It can be confusing when user will want to have CUPS error_log in syslog, but others in /var/log/cups directory - missing one or two 'View X log' buttons can look bad when the rest will be there.

What is your opinion on these two points? Is it acceptable to read cups-files.conf, set new cgi variable, change admin template a little and set "INPUT TYPE={?variable}" according cgi variable (SUBMIT/HIDDEN)?

michaelrsweet:
Really, given the current (default) use of systemd logging on most Linux distros and the disabling of the web interface by default, my personal preference is just to remove the buttons completely from the admin page.

zdohnal:
Ok, I'll prepare the removal of log buttons from web ui (even when /var/log/cups location is used).

michaelrsweet:
[master 0d94b8bf9] Remove log buttons from admin page since most Linux distros use systemd logging (Issue #5513)

PR #5519, opened 2019-02-18
Make explicit cast from enum to integer

IslandCow:
Clang-7 complains that there is a cast from unsigned (ipp_finishing_t which is an enum) to signed (int).  Perform an explicit cast since this should always be safe as long as ipp_finsihing_t is not too large.


michaelrsweet:
OK, sure.

PR #5528, opened 2019-02-24
Harden cupsd systemd service by default

paulmenzel:


paulmenzel:
I got an email from Travis CI, saying the build [failed](https://travis-ci.org/paulmenzel/cups/builds/497729103?utm_medium=notification&utm_source=email).

The problem is unrelated to my change though.

```
E: Unable to locate package libsystemd-dev
The command "sudo apt-get install -y libavahi-client-dev libgnutls28-dev libkrb5-dev libnss-mdns libpam-dev libsystemd-dev libusb-1.0-0-dev zlib1g-dev" failed and exited with 100 during
```

No idea, why it can’t be found, as the package [exists in all Ubuntu releases](https://packages.ubuntu.com/search?keywords=libsystemd-dev).

michaelrsweet:
@paulmenzel How widely implemented are these security features?

Also, at first glance we'll need to use `@` variables for the writable directories, not to mention that a *lot* of drivers depend on other locations being writable as well.

paulmenzel:
On 02/25/19 05:06, Michael R Sweet wrote:
> @paulmenzel How widely implemented are these security features?

Do you mean in general or just for the CUPS daemon? In general,
they are now widely implement, and a lot of services use them. For
CUPS it’s not widely used yet.

> Also, at first glance we'll need to use `@` variables for the
> writable directories,

Sorry, I missed that, and have now amended the commit. Is
`@CUPS_SERVERROOT@` meant for `/etc/cups`?

> not to mention that a *lot* of drivers depend on other locations
> being writable as well.
In separate paths, or below the CUPS directories?

michaelrsweet:
@paulmenzel I was more concerned with whether these changes were widely implements/supported by systemd - if this is a relatively new thing I'd rather have the distros pick this up first and only incorporate it once things have settled down.

`@CUPS_SERVERROOT@` is for /etc/cups, and the other directories each have a corresponding variable.

As for writable directories, yes there are a number of drivers and third-party solutions that regularly write outside the usual CUPS directories. For macOS we use more of a blacklist approach for system directories, and apply custom sandbox profiles to print filters and other "helper" programs separate from the scheduler (cupsd). Using the same security profile for cupsd and its children is of limited usefulness... :/

CameronNemo:
FWIW Ubuntu has shipped cups with an AppArmor profile for years. Debian will do the same for the upcoming Buster release. Perhaps LSM policy is a more fruitful avenue to pursue, considering how limited systemd's security directives can be.

Maryse47:
@CameronNemo only few distros have AppArmor enabled by default while most distros use systemd. Also systemd  security directives aren't limited in comparison with AppArmor, actually it's the opposite  - systemd can handle seccomp rules while AppArmor doesn't.

CameronNemo:
the only seccomp feature you are using is the w^x directive. that can easily be emulated with apparmor by only allowing files to be mapped or written, not both.

Maryse47:
@CameronNemo 

No, systemd allows for blacklist/whitelist specific syscalls, not used here but possible.

Also AppArmor doesn't allow to express w^x for the same file which is the point.

I don't know why you refer to "you are using" when I didn't created this PR.

michaelrsweet:

PR #5529, opened 2019-02-24
Optimise graphics (PNG+PDF) with zopflipng and ghostscript

Rigo-V:
Shaved 1.8MiB off the pdf files test/*.pdf.


Rigo-V:
This seems to contain the PNG optimisations from the pull request [5527](https://github.com/apple/cups/pull/5527) as well, so I closed that one.

michaelrsweet:
Since these files are generated using a separate program, and since Ghostscript-generated PDFs are not always the nicest, I hesitate to accept this pull request. Will investigate incorporating zopflipng (or something like it) into the PNG generation toolchain, but Ghostscript is a non-starter for this.

Rigo-V:
I understand your hesitation regarding the use of ghostscript. It sometimes make ugly pdfs. I used this command:
`gs -sDEVICE=pdfwrite -dBATCH -dNOPAUSE -dPDFSETTINGS=/prepress -dCompatibilityLevel=1.4 -sOutputFile=file.pdf file.pdf.uncompressed`
for compressing onepage-a4.pdf, onepage-letter.pdf, document-a4.pdf and document-letter.pdf, and cannot notice any loss of detail in the compressed files.
I cannot remember which exact command I used for the testfile.pdf, but there seems to be no degradation in it either (apart from the non-functioning hyperlinks, but that shouldn't be much of a problem since the file is for testing printing, right?

It'd obviously be fantastic if you incorporated optimisations in your toolchain itself. If that happens I definitely wouldn't mind this PR be rejected. Zopflipng seems like the best but also the slowest. But since the PNGs seem to be changed very rarely, this shouldn't be much of an issue.
I use the following options myself:
`--iterations=500 --filters=01234mepb --lossy_transparent`

michaelrsweet:
OK, I've done a bunch of testing and am happy with the results. Changes are getting pushed now...

michaelrsweet:
OK, changes are now merged.

PR #5531, opened 2019-02-25
Fix Travis CI configuration

paulmenzel:


michaelrsweet:
Sorry, I'm in the process of removing the travis support from the mainline CUPS due to my lack of privileges to grant repository access to Travis CI. I wish this was not so, but ...

paulmenzel:
I’d vote for keeping it even though you cannot directly use it, I still can, and it already found some things.

This pull request is still work in progress, as I put a commit trying to fix a build issue on top.

PR #5535, opened 2019-02-27
Fix saving cupsJobPassword in PPD cache

Nikitosh:
It seems there are typos in saving/reading PPD cache file, therefore "job-password-supported" attribute can't be fetched correctly.


michaelrsweet:
OK, so the proposed changes are exactly correct, and it looks like there are also issues with JobAccountingId and JobAccountUserId...

Will be pushing a slightly different set of changes shortly...

michaelrsweet:
[master d88008a0e] Fix PPD caching and IPP Everywhere PPDs (Issue #5535)

[branch-2.2 dc792541e] Fix PPD caching and IPP Everywhere PPDs (Issue #5535)

Issue #5550, opened 2019-03-28
Race condition on boot between cups and sssd


vtapia:
When cups has set the "SystemGroup" directive to an external group provided through sss and cups starts before sssd has finished booting, cups will crash because the group does not exist. For instance, with a group named lpadmins@tests.local served from Active Directory through sssd, if the sssd service hasn't booted before cups:

Mar 27 10:10:33 cups-sssd systemd[1]: Started CUPS Scheduler.
Mar 27 10:10:33 cups-sssd systemd[1]: Started CUPS Scheduler.
Mar 27 10:10:33 cups-sssd systemd[1]: Started Make remote CUPS printers available locally.
Mar 27 10:10:33 cups-sssd cupsd[21463]: Unknown SystemGroup "lpadmins@tests.local" on line 19 of /etc/cups/cups-files.conf.
Mar 27 10:10:33 cups-sssd cupsd[21463]: Unable to read "/etc/cups/cups-files.conf" due to errors.
Mar 27 10:10:33 cups-sssd systemd[1]: cups.service: Main process exited, code=exited, status=1/FAILURE
Mar 27 10:10:33 cups-sssd systemd[1]: cups.service: Failed with result 'exit-code'.
Mar 27 10:10:33 cups-sssd systemd[1]: cups.service: Service hold-off time over, scheduling restart.
Mar 27 10:10:33 cups-sssd systemd[1]: cups.service: Scheduled restart job, restart counter is at 2.
Mar 27 10:10:33 cups-sssd systemd[1]: Stopping Make remote CUPS printers available locally...
Mar 27 10:10:33 cups-sssd systemd[1]: Stopped Make remote CUPS printers available locally.
Mar 27 10:10:33 cups-sssd systemd[1]: Stopped CUPS Scheduler.

If sssd is available before cups starts, everything works as expected.

vtapia:
Adding "After=sssd.service" to the [Unit] section in scheduler/org.cups.cupsd.service.in should be enough to fix this issue. I'll prepare a PR with the change.

michaelrsweet:
Closing this issue; will track changes in the referenced pull request instead.

PR #5551, opened 2019-03-28
scheduler/org.cups.cupsd.service.in: ensure that sssd is up when cups starts

vtapia:
When the group defined in the SystemGroup directive is provided via
sssd, and sssd is not available when cups starts, cups will crash. This
patch makes cups start after sssd when both are present in the machine.
(Issue #5550)

### DIFF
--- a/scheduler/org.cups.cupsd.service.in
+++ b/scheduler/org.cups.cupsd.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=CUPS Scheduler
 Documentation=man:cupsd(8)
+After=sssd.service
 
 [Service]
 ExecStart=@sbindir@/cupsd -l


imsedgar:
If there is a After= dependency to the main repository, then the default from Fedora should also included:

After=network.target ypbind.service

CUPS requires a working network connection (to reach network printers), and sssd is just one possible name service, so ypbind should also be included. I would request to change the line to

After=network.target ypbind.service sssd.service

zdohnal:
@imsedgar Thanks for raising the issue up! I added that patch in Fedora some time ago, but I was not sure if it is a issue for upstream. The patch is for users, which use NIS for getting users/groups from other machine and put NIS user/group into SystemGroup. (But since NIS is quite old and not very used, I was not sure about reporting upstream).

imsedgar:
@zdohnal In general I think that only "After=network.target" is necessary, because cups is a service that needs only access local files and the network. Packages (rpm, deb) can create a local user for this local service. This service is independent from the same service on other computers, there is no need to use the same uid, gid, username or group name between different hosts. So I think it is preferable not to share the user and group between different hosts, but keep it local to each host. Because of this I think it is a bad idea to share it in nis, ldap, active directory, etc.

For local use, I think, if sssd is configured but not running, then the name resolution interfaces will access files in /etc directly.

michaelrsweet:
[master 4d0f1959a] Update systemd service file for cupsd (Issue #5551)

[branch-2.2 aaebca566] Update systemd service file for cupsd (Issue #5551)

PR #5557, opened 2019-04-05
Fix cupsCheckDestSupported() handling of "job-password" option

Nikitosh:
CUPS doesn't handle correctly "job-password" option in cupsCheckDestSupported() function, since it expects the integer value when IPP tag of "X-supported" attribute is IPP_TAG_INTEGER, but in case of "job-password" we have string value.
This commit adds separate check for that and fixes it.

### DIFF
--- a/cups/dest-options.c
+++ b/cups/dest-options.c
@@ -264,6 +264,15 @@ cupsCheckDestSupported(
         pwg->length >= min_length && pwg->length <= max_length)
       return (1);
   }
+  else if (!strcmp(option, "job-password")) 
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Hmm, I think we want to make this more general. Investigating.

michaelrsweet:
[master 677e4ca62] For octetString values, compare the length against the -supported value (Issue #5557)

[branch-2.2 0ad7895cf] For octetString values, compare the length against the -supported value (Issue #5557)

PR #5558, opened 2019-04-05
Fix handling of OctetString in get_options() function

Nikitosh:
The OctetString option was handled as if they have IPP_TAG_NAME tag, though it stores its data in unknown.data instead of string.text and also stores length instead of '\0' end symbol.


michaelrsweet:
This fix is incomplete - ipp_length() also needs to be updated...

michaelrsweet:
I'm also opting to support hex encoding when the string contains binary data.

[master 5483fd294] Correctly encode octetString values for print filters (Issue #5558)

[branch-2.2 803b6de89] Correctly encode octetString values for print filters (Issue #5558)

Issue #5576, opened 2019-05-07
2.3b8: 4.4-subscription-ops.test fails with server-error-internal-error (Invalid group tag.)


OdyX:
Thanks for the 2.3b8 version.

Unfortunately, in a Debian `sid` chroot, the `4.4-subscription-ops.test` fails with the following log:

```
[07/May/2019:08:05:00 +0000] "4.4-subscription-ops.test":
    Add Printer Subscription w/Lease                                     [PASS]
        notify-subscription-id (integer) = 3
    Verify Subscription Expiration                                       [PASS]
    Add 2 Printer Subscriptions w/Lease                                  [FAIL]
        RECEIVED: 0 bytes in response
        status-code = server-error-internal-error (Invalid group tag.)
        IPP request failed with status server-error-internal-error (Invalid group tag.)
    List Printer Subscriptions                                           [PASS]
        notify-events (keyword) = printer-state-changed
        notify-printer-uri (uri) = ipp://localhost:8631/printers/Test1
        notify-subscription-id (integer) = 4
    Check MaxSubscriptions limits                                        [PASS]
```

I can reliably reproduce it with the following minimal build-script:
```bash
#!/bin/sh
set -e
export LANG=C
aclocal
autoconf
./configure --build=x86_64-linux-gnu \
        --prefix=/usr \
        --includedir=\${prefix}/include \
        --mandir=\${prefix}/share/man \
        --infodir=\${prefix}/share/info \
        --sysconfdir=/etc \
        --localstatedir=/var \
        --disable-silent-rules \
        --libdir=\${prefix}/lib/x86_64-linux-gnu \
        --libexecdir=\${prefix}/lib/x86_64-linux-gnu \
        --disable-maintainer-mode \
        --disable-dependency-tracking \
        --with-docdir=/usr/share/cups/doc-root \
        --localedir=/usr/share/cups/locale \
        --enable-libpaper \
        --enable-ssl \
        --enable-gnutls \
        --enable-threads \
        --enable-static \
        --enable-debug \
        --enable-dbus \
        --with-dbusdir=/etc/dbus-1 \
        --enable-gssapi \
        --enable-avahi \
        --disable-launchd \
        --with-cups-group=lp \
        --with-system-groups=lpadmin \
        --with-rundir=/run/cups \
        --with-printcap=/run/cups/printcap \
        --with-log-file-perm=0640 \
        --with-local_protocols='dnssd' \
        --with-systemd=/lib/systemd/system \
        --enable-libusb
[...]

PR #5577, opened 2019-05-08
Test: Fix *topdf filters to find the examples in the correct path

OdyX:
As identified in https://github.com/apple/cups/issues/5576#issuecomment-490148791, some files cannot be found in tests.

Indeed, the generated filters look for some example files in `test/` while they moved to `examples/` between v2.3b7 and v2.3b8.


michaelrsweet:
Ah, good catch!

PR #5591, opened 2019-06-07
insert explicit casts to calm compiler warnings for -Wsign-conversion

mpsuzuki:
During compilation of the latest source, I received several warnings for -WSign-conversion. Here is a patch inserting explicit casts to calm them.


michaelrsweet:
Sigh...  This is all because the newest GCC defaults to unsigned enums... Will consider for 2.3.x.

mpsuzuki:
I see. How about add "-Wno-sign-conversion" to CFLAGS as quick workaround?

michaelrsweet:
@mpsuzuki The problem with -Wno-sign-conversion is that it also eliminates valid issues, like between size_t and ssize_t, that we want to see reported.

michaelrsweet:
Merged mostly as-is...

[master 8ef4f507d] Eliminate new GCC compiler warnings (Issue #5591)

PR #5592, opened 2019-06-07
dump both of uri & securi if HAVE_SSL, dump uri only if not HAVE_SSL

mpsuzuki:
During compilation of the latest source, accidentally I found a failure in building without SSL. It seems that current source does not dump securi if SSL is available but tries to dump securi if SSL is unavailable. securi is not defined if SSL is unavailable, so I exchanged the debug messages for the cases with/without SSL.


mpsuzuki:
sorry, I found this PR is a duplicate of PR #5590, I close this.

PR #5593, opened 2019-06-07
remove unused to64() from scheduler/auth.c to calm warning for -Wunused-functions

mpsuzuki:
scheduler/auth.c has static to64() function if LIBPAM is unavailable, but there is no part using it. To calm -Wunused-functions, I removed to64(). Please could you consider?


mpsuzuki:
thank you for taking care on this!

PR #5594, opened 2019-06-10
Implement `cupsEnumDests2` function.

trialuser02:
I propose to implement the function cupsEnumDests2. It will allow to enumerate destinations using specified connection.


michaelrsweet:
Will consider, however one of the key elements of cupsEnumDests is that it doesn't depend on the application being able to talk directly to a printer - sandboxing can prevent outside network connections. By providing the http_t connection for the "local" server you run the risk of that breaking, and we don't want to provide more ways for developers to get things wrong...

trialuser02:
But other functions have alternative with http_t* argument (cupsGetDests -> cupsGetDests2). It is rather strange that this function is not in this list.

michaelrsweet:
@trialuser02 cupsEnumDests provides an abstraction for the connection details - you discover available printers with cupsEnumDests, and connect to the appropriate service with the cupsConnectDest API afterwards.

michaelrsweet:
Sorry, we've decided not to implement this function.

PR #5606, opened 2019-07-03
Fix "sides" option if using *Duplex* in lpoptions file

roberthoffmann:
Remove *Duplex* option read from lpoptions file if "sides" is
specified on command line. Otherwise both options would be
present and "sides" would not override Duplex.

Test case:
Have a ~/.cups/lpoptions file containing a Duplex option like this:
$ cat ~/.cups/lpoptions
Default myprinter Duplex=None PageSize=A4

Print a document with more than one page on a printer supporting
duplex:
$ lpr -P myprinter -o sides=two-sided-long-edge test.pdf

Without this patch the document is not printed two-sided,
with this patch applied it is printed two-sided.

### DIFF
--- a/cups/options.c
+++ b/cups/options.c
@@ -80,6 +80,15 @@ cupsAddOption(const char    *name,	/* I  - Name of option */
   else if (!_cups_strcasecmp(name, "print-quality"))
     num_options = cupsRemoveOption("cupsPrintQuality", num_options, options);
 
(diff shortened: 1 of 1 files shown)


michaelrsweet:
This isn't strictly the correct fix, as there are many PPD options that map to sides... Will consider for a future update.

michaelrsweet:

PR #5613, opened 2019-07-16
Lookup groups using group->gr_mem even if getgrouplist() failed

manu0401:
getgrouplist() iterate on each group, searching if the user
is included, while group->gr_mem iterated on user list for
the target group. In some case the former method can miss
the result while the later will succeed. This is the case
for OpenLDAP dyngroup, where the group members are automatically
generated on group search. Searching by group member gives no
result, while searching the member of a given groups is fine.

### DIFF
--- a/scheduler/auth.c
+++ b/scheduler/auth.c
@@ -1188,13 +1188,19 @@ cupsdCheckGroup(
 	  return (1);
     }
 
-#else
+#endif /* HAVE_GETGROUPLIST */
+    /*
+     * Searching group->gr_mem after getgrouplist failure may
+     * yield success because some NSS source have better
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Will consider this, although I am definitely not super happy about the inconsistency...

michaelrsweet:
I reworked this change to move the membership list check up front, followed by the getgrouplist check. We already have the list of usernames that are members, so this made more sense than checking after the (semi-expensive) getgrouplist check.

[master 3c27d2a6d] Always check the group membership list (Issue #5613)

[branch-2.2 666d07699] Always check the group membership list (Issue #5613)

PR #5620, opened 2019-07-26
Introduce cupsHashNoCryptoData and --enable-gnutls-relax-mode configu…

zdohnal:
…re option

The cupsHashNoCryptoData function is for hashing non cryptographic data by MD5
hash function. MD5 hash function can be disabled in GNU TLS on specific machines
so new --enable-gnutls-relax-mode enables GNU TLS relax mode for use cases of MD5
which are not considered as encryption.


zdohnal:
I check it for cookie and UUID creation and it works.

michaelrsweet:
No

Issue #5601, opened 2019-06-19
Work around GNU TLS hard deprecation of MTI hash functions


zdohnal:
Hi,

I did further cryptographic review of packages which we ship in our operating systems regarding security certifications - e.g. FIPS - and I found out that there are several usages of MD5 algorithm, which is not considered to be safe according to FIPS certification and current security standards. Additionally Fedora and Red Hat Enterprise Linux are moving towards the ability to set the crypto policy on system-wide level [https://fedoraproject.org/wiki/Changes/CryptoPolicy] and MD5 is not allowed in the default settings/profile.

I plan to create a patch for replacing MD5 with a more secure algorithms for cryptographic use cases plus relaxing gnutls for non cryptographic use cases e.g. generating UUID and cookie and I would like to get it to the upstream sources, but I am not completely sure whether this is something that the project could be interested in and potentially accept? Additionaly if you have any guidance or advices to make the patch more acceptable for the project, please let me know.

Using secure and more recent crypto algorithms is beneficial not only for security certifications, but also in general, therefore I hope that you will consider changes that I would like to propose.

Thank you for your time and have a nice day,

Zdenek

michaelrsweet:
@zdohnal What uses of MD5 are you concerned about in CUPS?

AFAIK, none of the remaining uses of MD5 matter:

- MD5 for Digest authentication is a requirement unless the server/printer supports SHA2 (which we detect and use automatically)
- MD5 for UUID generation is OK because there is no way to do prefix attacks and we don't use a lot of bits anyways
- MD5 for showing a digest of a certificate chain to the user is purely informational - the actual validation compares the contents (bytes) of the certificate chain and not the hashes

MD5 is never used with TLS, and MD5 as a hash algorithm for certificates is no longer allowed either.

zdohnal:
> @zdohnal What uses of MD5 are you concerned about in CUPS?
> 
> AFAIK, none of the remaining uses of MD5 matter:
> 
>     * MD5 for Digest authentication is a requirement unless the server/printer supports SHA2 (which we detect and use automatically)

You're right, I missed the first if block, where SHA2 can be used.

> 
>     * MD5 for UUID generation is OK because there is no way to do prefix attacks and we don't use a lot of bits anyways
> 

Seems good to me too.

>     * MD5 for showing a digest of a certificate chain to the user is purely informational - the actual validation compares the contents (bytes) of the certificate chain and not the hashes
> 

[...]

PR #5622, opened 2019-07-29
Introduce --enable-gnutls-relax-mode, apply it in cupsHashData()

zdohnal:
#5601

### DIFF
--- a/config-scripts/cups-ssl.m4
+++ b/config-scripts/cups-ssl.m4
@@ -10,6 +10,7 @@ dnl
 AC_ARG_ENABLE(ssl, [  --disable-ssl           disable SSL/TLS support])
 AC_ARG_ENABLE(cdsassl, [  --enable-cdsassl        use CDSA for SSL/TLS support, default=first])
 AC_ARG_ENABLE(gnutls, [  --enable-gnutls         use GNU TLS for SSL/TLS support, default=second])
+AC_ARG_ENABLE(gnutls_relax_mode, [  --enable-gnutls-relax-mode         use GNU TLS in relax mode for MD5 hash function at non-crypto cases, default=disabled])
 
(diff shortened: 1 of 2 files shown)


michaelrsweet:
@zdohnal Is the GNUTLS_FIPS140_SET_LAX_MODE call not an actual function? I would prefer to just auto-detect it at configure time and call it if it is defined. The more knobs we get the less likely it is that it will work... :/

zdohnal:
It is actually macro for:

if (gnutls_fips140_mode_enabled())
  gnutls_fips140_set_mode(GNUTLS_FIPS140_LAX, GNUTLS_FIPS140_SET_MODE_THREAD);

See https://gnutls.org/manual/html_node/FIPS140_002d2-mode.html

michaelrsweet:
@zdohnal Based on the documentation for that function, we can safely do a configure check for gnutls_fips140_set_mode and then do:

```
unsigned oldmode = gnutls_fips140_mode_enabled();
gnutls_fips140_set_mode(GNUTLS_FIPS140_LAX, GNUTLS_FIPS140_SET_MODE_THREAD);

...

gnutls_fips140_set_mode(oldmode, GNUTLS_FIPS140_SET_MODE_THREAD);
```

This will save and restore the mode around the hashing calls, and hopefully not step on any other users of this API (since we'd restore the previous setting, not just put it back to strict...)

michaelrsweet:
Pushed changes for both 2.3.x and 2.2.x. Please review and let me know if this addresses your issues...

[master 9ec6124c3] GNU TLS FIPS140 support (Issue #5601, Issue #5622)

[branch-2.2 2f26c6b71] GNU TLS FIPS140 support (Issue #5601, Issue #5622)

PR #5611, opened 2019-07-10
Add extra NULL check before calling strlcpy().

j39m:
We detected a crash in `_ppdCreateFromIPP()` when calling `[_cups_]strlcpy()`. The trace suggested that the `ippGetString()` call was returning `NULL`.

Unfortunately, I was unable to reproduce this crash locally.

PR #5635, opened 2019-08-16
Add extra NULL check before calling strlcpy().

j39m:
This pull request is a copy of pull #5611, which I closed for bad git hygiene.


michaelrsweet:
@j39m This is still not something I want to change - ippGetString *cannot* return NULL under these conditions, so either the ChromeOS fork of CUPS has broken something or there is some other memory corruption happening that is causing this behavior.

We really need to be able to reproduce this before any action can be taken. Is there nothing you can provide to further isolate the cause?

j39m:
I'll try to reproduce this and will close this PR if nothing turns up. Sorry for the trouble.

j39m:
I tried hacking at `ippserver` to reproduce the crash but turned up empty-handed.

PR #5649, opened 2019-09-13
Create untranslated cups.da.po

scootergrisen:
It seems Debian currently only includes the other translated files (like templates) if this file exists.
So i create this untranslated cups.da.po file and hope Debian will include danish translation in their CUPS packages.


scootergrisen:
Should have been cups_da.po not cups.da.po.

PR #5650, opened 2019-09-13
Create untranslated cups_da.po

scootergrisen:
It seems translated files (templates etc.) is only included in CUPS packages for distributions if a po file exists for that language.
So i create this untranslated cups_da.po file and hope danish translation will be included.

### DIFF
--- a/locale/cups_da.po
+++ b/locale/cups_da.po
@@ -1,0 +1,18927 @@
+# Danish translation for CUPS.
+# Copyright © 2007-2019 by Apple Inc.
+# Copyright © 2005-2007 by Easy Software Products.
+# Licensed under Apache License v2.0.  See the file "LICENSE" for more
+# information.
+# Notes for Translators:
+# The "checkpo" program located in the "locale" source directory can be used
+# to verify that your translations do not introduce formatting errors or other
+# problems.  Run with:
+# cd locale
+# ./checkpo cups_LL.po
+# where "LL" is your locale.
+# scootergrisen, 2019.
+msgid ""
+msgstr ""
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Sorry, until we have a significant number of translated strings we can't include it...

scootergrisen:
How many is "significant number"?

michaelrsweet:
We'd want many of the printer-state-reasons keywords (marker-xxx, media-xxx, and toner-xxx are the most common) along with the messages from the common print commands (cancel, lp, lpadmin, lpc, lpoptions, lpq, lpr, lpstat) covered, at least.

scootergrisen:
It seems no language have translated marker-*, media-* and toner-*.
Expect "en" but that's the source language.

And its not always clear what the source string is/what i should translate.
So i think it would be better to fix this before asking people to translate it.
If i translate it now and the following problem is fixed i will have to translate it once again.

If cups_da.po says:

> #. TRANSLATORS: Marker Opc Missing
> msgid "printer-state-reasons.marker-opc-missing"
> msgstr ""

First of all the string i should translate should be in msgid.
But ok i look at the comment instead and use "Marker Opc Missing" as the source string.

But then if i look in cups_en.po the string that is suppose to be the source string is not the same.
Its "OPC missing" not "Marker Opc Missing" like the comment says.

> #. TRANSLATORS: Marker Opc Missing
> msgid "printer-state-reasons.marker-opc-missing"
> msgstr "OPC missing"

scootergrisen:
I made https://github.com/apple/cups/pull/5707 instead.

PR #5663, opened 2019-10-10
Pessimize PPD choice allocation

j39m:
We noticed that our Clusterfuzz instance occasionally [had trouble allocating `ppd_choice_t`s once the test cases got huge](https://bugs.chromium.org/p/chromium/issues/detail?id=992067). This change aims to address that by allocating memory a little more greedily.

### DIFF
--- a/cups/ppd.c
+++ b/cups/ppd.c
@@ -61,6 +61,7 @@ static pthread_once_t	ppd_globals_key_once = PTHREAD_ONCE_INIT;
 static ppd_attr_t	*ppd_add_attr(ppd_file_t *ppd, const char *name,
 			              const char *spec, const char *text,
 				      const char *value);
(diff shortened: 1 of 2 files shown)


michaelrsweet:
Sorry, but we will not accept this pull request for a few reasons:

1. The current code is optimized for the most common usage, and optimizing with power-of-two increases will dramatically increase the memory usage of CUPS which is not what we want.

2. The proposed changes break binary compatibility with the current PPD API by changing the size of (unfortunately) public structures.

3. The PPD API has been deprecated for a very long time and we have no interest in optimizing or making major changes to it at this point in time.

4. PPD-based printer drivers are deprecated in CUPS 2.3, and the preferred replacement has been  IPP Everywhere since CUPS 1.4.

PR #5679, opened 2019-11-08
Make printer instances work again

michaelweghorn:
Hi Mike,

since commit c5f5c5a7683924618b3e6a0ab6db5a1674db8f6d, printer instances don't work anymore.

They can be created, e.g. using 

    lpoptions -p myprinter/myinstance -o Duplex=None

and then show up in `~/.cups/lpoptions`. However, they are not listed when e.g. running `lpstat -v` or in print dialogs.

These two commits make printer instances work again for the case where queues from the local CUPS server are used.

Handling of instances for autodiscovered printers might need a follow-up commit. I wasn't 100 % sure about this, so left it unchanged.

### DIFF
--- a/cups/dest.c
+++ b/cups/dest.c
@@ -3457,12 +3457,6 @@ cups_enum_dests(
 
   DEBUG_printf(("1cups_enum_dests: def_name=\"%s\", def_instance=\"%s\"", data.def_name, data.def_instance));
 
-  /*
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Sorry, that change has some bad side-effects. I will push a simpler fix to cups_enum_dests against the parent bug...

michaelweghorn:
Fine for me, thanks for looking into it!

PR #5690, opened 2019-11-23
treewide: Replace bzero with memset

neheb:
bzero has been deprecated by POSIX 2008. It recommends the use of memset
instead.


michaelrsweet:
It may be deprecated by POSIX, but this is platform specific code and zero is not deprecated on macOS. Will consider it...

neheb:
OTOH that Darwin file already uses mender.

PR #5695, opened 2019-12-04
Off by one error in ipp_finishings_vendor

IslandCow:
When enumvalue is 101 and attrname is "finsishings-supported"
we were getting the memory after ipp_finishings_vendor
in the ipp_job_collation_types array.


michaelrsweet:
I don't see how this happens with value 101. 0x40000066 maybe, but not 101.

Anyways, will take the fix.

IslandCow:
Thanks for pulling.  The problematic value is 0x40000065 because the length of the array is 0x65 or 101.

PR #5710, opened 2019-12-27
Fix permissions of ipp Backend for legacy driverless printing support

logTom:
Cups-filters 1.26.0+ assumes 755 for the driverless printing fallback logic.
Ubuntu 19.10 already applies a distro specific patch to cups for this.

The issue was brought up here
https://github.com/OpenPrinting/cups-filters/issues/163.

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -13,7 +13,8 @@ include ../Makedefs
 # Object files...
 #
 
-# RBACKENDS are installed mode 0700 so cupsd will run them as root...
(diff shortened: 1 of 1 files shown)


michaelrsweet:
That would break Kerberized printing, can't do it.

logTom:
Thank you for your feedback.
I created a new issues for cups-filters to discuss how we can fix this otherwise.
https://github.com/OpenPrinting/cups-filters/issues/183

PR #5722, opened 2020-01-20
Prioritize print-color-mode-supported over PWG when creating color mo…

### LINKED ISSUE #5713: 'everywhere' model does not create default color models in created PPD when IPP printer response is insufficient
Hi,

I tried it only with my HP LaserJet printer (monochrome) at home and with Canon ImageRunner (color and monochrome) in our office, but ppd generator did not create ppd with values similar to https://www.cups.org/doc/ppd-compiler.html#COLOR .
This issue causes the user is not able to select color mode in print dialog.
Here is the ppd of canon printer created by 'lpadmin -p printer -v uri -m everywhere -E' 
[canon.ppd.txt](https://github.com/apple/cups/files/4018470/canon.ppd.txt) and result of get-printer-attributes request 
[get-printer-attributes.log.txt](https://github.com/apple/cups/files/4018472/get-printer-attributes.log.txt) .

zdohnal:
…dels

The same patch as for cups-filters ppd generator, fixes #5713 .


zdohnal:

Issue #5511, opened 2019-02-06
PPD retrieval issues with MacOS Mojave 10.14.(3?)


kabassanov:
Hello,

We actually experience some (random) issues with Apple computers when we try to install printers shared by Bonjour.

Automatic driver selection often fails to "Generic PostScript Printer" instead of the right one, and according to tcpdump  captures, it seems to be related to the presence of a field called "If-Modified-Since" in the get PPD cups request like: 
If-Modified-Since: Fri, 28 Oct 432601864 03:13:58 GMT\r\n

User-Agent: CUPS/2.2.9 (Darwin 18.2.0; x86_64) IPP/2.0\r\n

Sometimes the field is absent in the packet and driver selection succeeds, sometimes (not really sure about this) it succeeds for some particular values of the year… supposing that 432601864  is a year.

Any hint?

Thanks.

michaelrsweet:
Please file a bug report directly with Apple: <https://bugreport.apple.com>

The 432601864 is indeed a year, but it isn't clear to me where it would be getting such a large date/time value.

efuss:
Until Apple fixes the GUI, you can safely ignore these garbage dates in the server because RFC 7231 says the year field in a HTTP-date it of type 4DIGIT.

[This patch](https://github.com/apple/cups/files/3749724/year-4DIGIT.patch.txt) made our clients work again.

efuss:
Since it looks I cannot re-open this issue, I created Pull request #5724 for the work-around.

PR #5724, opened 2020-01-21
Ignore HTTP-date values with invalid year field

efuss:
Some clients request PPD files with a random value (e.g.., 25 Nov 4461684 22:08:41) in the `If-Modified-Since` header (see Issue #5511).
Ignore this by rejecting invalid (RFC 7231 says the year field is `4DIGIT`) years in `httpGetDateTime()`.


michaelrsweet:
Fixed in OpenPrinting CUPS.

PR #5738, opened 2020-02-14
Fix leakage of ppd

zdohnal:
Hi,

I run our internal RH coverity scan, which identified memory leak of ppd structure.

Would you mind merging it into the project?


zdohnal:

PR #5770, opened 2020-04-20
Fix inability to differentiate between multiple same-model printers on USB (dymo)

inflex:
Problem: 
cups will not properly differentiate between multiple same-model devices on the USB bus if they do not have a unique serial number in the IEEE1284 ID string, even if the USB data does in fact have a unique serial attribute.

Modification/Fix:
Changed make_device_uri() so that it will look for a serial number in the USB device data first before it tries to find the data in the IEEE1284 string.

Existing code also was possibly not likely to work because the devdesc struct was not initialised with libusb_get_device_descriptor() before _get'ing the iSerialNumber.

Code now permits the use of multiple printers of the same model/type even with the same IEEE1284 data SERN which was the case with the DYMO LabelWriter 450 units.

### DIFF
--- a/backend/usb-libusb.c
+++ b/backend/usb-libusb.c
@@ -1251,7 +1251,7 @@ make_device_uri(
     char          *uri,			/* I - Device URI buffer */
     size_t        uri_size)		/* I - Size of device URI buffer */
 {
-  struct libusb_device_descriptor devdesc;
+  struct libusb_device_descriptor devdesc; /* Holds USB device information after loaded */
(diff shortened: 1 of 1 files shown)


inflex:
Just a heads-up that I thought about last night;   this fix may have some problematic side-effects in that the DeviceURI could potentially change for existing installed systems if they didn't previously have a serial number in the DeviceURI and subsequently could result in existing installations not properly functioning after an update.

michaelrsweet:

PR #5773, opened 2020-04-26
fix wide roll printing: In rastertopwg.c, don't core dump if map.pwg is NULL

pacew:
I don't have any understanding of the big picture, but after I recently updated my ubuntu system, I could no longer print to my HP T120 wide format roll printer.  I tracked it down to a core dump from rastertopwg caused by an attempt to reference pwg_size->map.pwg when it was NULL.  With this change in place, I can print to my T120 again.


michaelrsweet:
Fixed in OpenPrinting CUPS.

Issue #5757, opened 2020-03-20
ipp backend doesn't consider contimeout value set via DEVICE_URI


mstefan-spr:
Hi all,

I am using CUPS 2.2.6 on RHEL 8.0 . If I pass an "unreachable" IP via an IPP URI, then then the job seem to stay forever in processing.  I found out there is a default connection timeout set to 1 week (604800 seconds) for the ipp backend so I'v tried using 30 seconds for the contimeout argument . Even if I pass the 'contimeout' argument it still loops forever trying to reach the printer.
This is how I set DVICE_URI: ipp://100.100.100.100:631/printer?contimeout=30

Let me know if other info is needed. 

Regards,
Marius

zdohnal:
Hi @mstefan-spr ,

I encountered the same issue with the newest cups-2.3.1- it happens because httpAddrConnect2 sets errno to EINPROGRESS after the first failure on connect(), then keeps reconnecting till timeout is reached, but it never sets errno to different value after that.

Then any backend which called connect() function in a way (directly(lpd)/via httpReconnect() (ipp)/ via httpAddrConnect() (socket)) gets EINPROGRESS which is not handled specially and contimeout is not applied at all, resulting in the job blocks the queue for eternity.

@salgernon  I have the following proposition:
1. add EINPROGRESS into handled errnos in every network backend (lpd,socket,ipp)
2. assign a different errno (e.g. ETIMEDOUT) in httpAddrConnect2() when timeout is reached and execution goes out of loop - this way will not fix the issue for lpd, where direct connect() is used, so EINPROGRESS should be added in lpd either way.

jsmeix:
Only as a side note FYI related to connect timeout versus
possible subsequent hangup (i.e. after connect succeeded)
see https://github.com/apple/cups/issues/5559

I.e. fixing only connect issues is not sufficient
to avioid that jobs may stay forever in processing state.

michaelrsweet:
So right now it looks like contimeout is supported in CUPS 2.3.x. If you are still having problems with the current version of CUPS, please report this to the [OpenPrinting CUPS](https://github.com/openprinting/cups) project.

PR #5782, opened 2020-05-21
Generate ETIMEDOUT when remaining_time is reached and EHOSTDOWN when …

zdohnal:
…all addresses are tried

Hi,

the patch fixes issue #5757 , where httpAddrConnect2() sets 2 other errnos - ETIMEDOUT and EHOSTDOWN - when two conditions are met:

- when remaining_time is reached, it sets errno to ETIMEDOUT
- when there are no other addresses and number of file descriptors is zero, it sets errno to EHOSTDOWN

This will propagate error into backends, where contimeout can be applied now. Other usages httpAddrConnect2() will not change:

1) httpAddrConnect() - is used in socket backend (where previous EINPROGRESS isn't taken care of and new errnos will result in applying correct timeouts), in cupsFileOpen() and ippfind tool, where the behavior will not change, because it checks only return value and not errno - httpAddrConnect2() still returns error as before
2) httpReconnect() - it will just return different errno in http->error - ETIMEDOUT or EHOSTDOWN instead of EINPROGRESS in those two mentioned cases. The previous result - the failure - will not change
3) ippeveprinter - still will return error.

Would you mind adding the patch to CUPS project? Please tell me if I can change anything.

Thank you in advance!

Zdenek


zdohnal:

PR #5789, opened 2020-06-06
HP Laserjet 1010 delay required

sgipower:
HP Laserjet suffers the same problem as the 1150 and 1015. The last page isn't printed until the job times out.


michaelrsweet:
Fixed in OpenPrinting CUPS

PR #5792, opened 2020-06-08
Web interface: Set the actual device uri, not device_uri|make_model s…

zdohnal:
…tring from html form

Hi @salgernon ,

if I try to install my printer with IPP everywhere model via CUPS web UI, it creates a raw print queue with incorrect device uri.

The reason is device_uri from html form template contains URI|make_model string - cgi script admin correctly gets make_model, but doesn't reset device_uri after that to the real device_uri.

Would you mind adding the patch to CUPS project?

Thank you in advance!

Zdenek


zdohnal:

PR #5793, opened 2020-06-11
ipptool: support mdns uris

zdohnal:
Hi,

ipptool utility doesn't support mdns uris, so debugging IPP printers is more difficult, because ```sudo lpinfo -l -v``` reports mdns uri for IPP print queues.

The patch does MDNS name resolution before doing ipp request.

Would you mind adding the patch into the project?

Thank you in advance,

Zdenek


zdohnal:

PR #5797, opened 2020-06-25
Update cups_de.po and clearify an expression in templates

jschwender:
added more translations (de) to po file. Using a more precise expression in the german templates.


michaelrsweet:
Fixed in OpenPrinting CUPS.

Issue #5808, opened 2020-07-21
Incorrect error message for non-existing printer when used in lp/lpr command


zdohnal:
Hi,

we have a report in [Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1811735) about incorrect message when you use a non-existing print queue for lp/lpr command - it is because ```cupsGetNamedDest()``` just returns NULL for a non-existing queue, it doesn't set any error and ```lpr``` nor ```lp``` check for ```IPP_STATUS_NOT_FOUND``` (looks to me as the most logical error).

I'll create a pull request with proposed fix.

michaelrsweet:
Fixed in OpenPrinting CUPS.

PR #5809, opened 2020-07-21
cups/dests.c: cupsGetNamedDest() - set IPP_STATUS_ERROR_NOT_FOUND err…

zdohnal:
…or if queue was not found

lp.c/lpr.c: check for IPP_STATUS_ERROR_NOT_FOUND and generate a proper message if hit

Hi,

here is the pull request connected to #5808 .

Localization files probably will need to be regenerated to support the new string in the pull request, but regenerating it by myself would cause conflicts for other people who has pull requests for localizations.
Please regenerate localization files whenever you see fit.


zdohnal:

PR #5823, opened 2020-08-14
Add USB quirk rule for Citizen CT-S4000 printer

nadrimajstor:
- Fix 8 seconds pause between print jobs.
- Fix (sometimes) fail to cut the receipt at the end.


michaelrsweet:
Fixed in OpenPrinting CUPS

PR #5827, opened 2020-09-07
backend/Makefile: Don't use libtool.

orbea:
This is not needed and causes problems with more strict implementations of libtool.

With slibtool it fails.
```
Installing backends in /tmp/package-cups/usr/lib64/cups/backend
rdlibtool: error: --mode must be specified.
rdlibtool: error: --mode must be specified.
make[1]: *** [Makefile:120: install-exec] Error 2
make: *** [Makefile:198: install-exec] Error 1
```
https://git.foss21.org/slibtool

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -118,7 +118,7 @@ install-exec:	$(INSTALLXPC)
 	echo Installing backends in $(SERVERBIN)/backend
 	$(INSTALL_DIR) -m 755 $(SERVERBIN)/backend
 	for file in $(RBACKENDS); do \
-		$(LIBTOOL) $(INSTALL_BIN) -m 700 $$file $(SERVERBIN)/backend; \
+		$(INSTALL_BIN) -m 700 $$file $(SERVERBIN)/backend; \
 	done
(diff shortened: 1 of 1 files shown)


prlw1:
FWIW this patch has been in http://pkgsrc.org since January 2016 (and a similar one since May 2010)
so can be considered tested in the wild on a variety of operating systems.

http://cdn.netbsd.org/pub/pkgsrc/current/pkgsrc/print/cups-base/patches/patch-af
(date not entirely obvious as cups package was renamed cups-base in 2017)

michaelrsweet:
Please submit this to the [OpenPrinting CUPS](https://github.com/openprinting/cups) project.

orbea:
@michaelrsweet Already done a while ago. https://github.com/OpenPrinting/cups/pull/11

Is this repo going to be finally replaced now?

michaelrsweet:
@orbea OpenPrinting has taken over CUPS development for Linux/BSD/etc. Apple is still developing CUPS for macOS. And I'm working to keep the two in sync...

PR #5831, opened 2020-09-21
Update org.cups.usb-quirks

AmadeusHF:
Fix for low speed USB raw data exchange due to Canon slow back-channel


michaelrsweet:
Fixed in OpenPrinting CUPS

PR #5835, opened 2020-10-03
Fix typo in ppdc documentation.

ferivoz:
PCFilenName is supposed to be PCFileName.


michaelrsweet:
Fixed in OpenPrinting CUPS.

PR #5836, opened 2020-10-06
Fix memory leaks

tsite:
http://crrev.com/c/2446885


michaelrsweet:
Fixed in OpenPrinting CUPS.

PR #5837, opened 2020-10-06
undefined behavior: signed integer overflow

tsite:


michaelrsweet:
Fixed in OpenPrinting CUPS.

PR #5846, opened 2020-10-19
Build ippfind in dnssd case, not just avahi and darwin cases.

prlw1:
This seems to be the intention of cups commit 766a822, but a comma
got in the way.


michaelrsweet:
Fixed in OpenPrinting CUPS.

PR #5951, opened 2021-08-30
Update m4, configure.ac, and configure

AtariDreams:
It's not perfect, but it works the same, and at least it does not require unneeded checks or ancient commands from older versions of autoconf


michaelrsweet:
Sorry, we aren't accepting large changes to the configure script for Apple CUPS, which is strictly in maintenance mode now.

PR #5952, opened 2021-08-30
Add _CUPS_NORETURN to missing functions

AtariDreams:
Almost every non-returning function has this qualifier, but not all. This should properly define them all as well as shrink the final binary a little bit.


michaelrsweet (changes requested):
See my comments to your OpenPrinting PR - the annotation should go on the prototype, not on the implementation of the function.

AtariDreams:
They already are though...

AtariDreams:
@michaelrsweet anything else?

PR #5953, opened 2021-09-03
Remove redundant code

AtariDreams:
Some of the code is either dead code, redundant, or always true. Meaning they do not do anything since the cases were already dealt with in earlier parts of the code. It would be best to remove this code.

Note that some of the code removal required a tiny bit of refactoring.

### DIFF
--- a/cups/auth.c
+++ b/cups/auth.c
@@ -330,10 +330,6 @@ _cupsSetNegotiateAuthString(
   gss_buffer_desc output_token = GSS_C_EMPTY_BUFFER;
 					/* Output token */
 
-
-  (void)method;
-  (void)resource;
-
 #  ifdef __APPLE__
  /*
   * If the weak-linked GSSAPI/Kerberos library is not present, don't try
@@ -450,6 +446,9 @@ _cupsSetNegotiateAuthString(
       }
     }
   }
+#  else
+  (void)method;
+  (void)resource;
 #  endif /* HAVE_GSS_ACQUIRED_CRED_EX_F */
 
   if (major_status == GSS_S_NO_CRED)
@@ -562,9 +561,9 @@ cups_auth_find(const char *www_authenticate,	/* I - Pointer into WWW-Authenticat
         * Skip quoted value...
         */
 
-        www_authenticate ++;
-        while (*www_authenticate && *www_authenticate != '\"')
+        do
           www_authenticate ++;
+        while (*www_authenticate && *www_authenticate != '\"');
 
         DEBUG_printf(("9cups_auth_find: After quoted: \"%s\"", www_authenticate));
       }
--- a/systemv/lpstat.c
+++ b/systemv/lpstat.c
@@ -634,7 +634,7 @@ match_list(const char *list,		/* I - List of names */
   if (!name)
     return (0);
 
-  while (*list)
+  do
   {
    /*
     * Skip leading whitespace and commas...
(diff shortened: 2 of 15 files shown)


cups/auth.c
```diff
@@ -330,24 +330,6 @@ _cupsSetNegotiateAuthString(
   gss_buffer_desc output_token = GSS_C_EMPTY_BUFFER;
 					/* Output token */
 
-
-  (void)method;
-  (void)resource;
-
-#  ifdef __APPLE__
```
michaelrsweet, line 337: This code cannot be removed as it is necessary for libcups to work on macOS when running without the Kerberos library (mainly in the installer).

scheduler/auth.c
```diff
@@ -724,20 +724,6 @@ cupsdAuthorize(cupsd_client_t *con)	/* I - Client connection */
 					/* Output token for username */
     gss_name_t		client_name;	/* Client name */
 
-
-#  ifdef __APPLE__
-   /*
-    * If the weak-linked GSSAPI/Kerberos library is not present, don't try
-    * to use it...
-    */
-
```
michaelrsweet, line 733: Same as for cups/auth.c. Can't be removed without breaking things on macOS.

scheduler/colorman.c
```diff
@@ -357,14 +357,6 @@ apple_register_profiles(
 			profile;	/* Current profile info dictionary */
   CFStringRef		dict_key;	/* Key in factory profile dictionary */
 
-
- /*
-  * Make sure ColorSync is available...
-  */
-
-  if (&ColorSyncRegisterDevice == NULL)
```
michaelrsweet, line 365: Ditto.

scheduler/conf.c
```diff
@@ -420,15 +420,6 @@ cupsdDefaultAuthType(void)
     return (default_auth_type);
 
 #ifdef HAVE_GSSAPI
-#  ifdef __APPLE__
- /*
-  * If the weak-linked GSSAPI/Kerberos library is not present, don't try
-  * to use it...
-  */
```
michaelrsweet, line 427: Ditto.

scheduler/ipp.c
```diff
@@ -2449,18 +2449,17 @@ add_printer(cupsd_client_t  *con,	/* I - Client connection */
...
-        cupsdDeletePrinter(printer, 0);
+    send_ipp_status(con, IPP_NOT_POSSIBLE, _("Bad port-monitor \"%s\"."),
+                    attr->values[0].string.text);
+    if (!modify)
+      cupsdDeletePrinter(printer, 0);
 
-      return;
-    }
+    return;
+
```
michaelrsweet, line 2461: Please don't introduce goto's for this. The only time we use goto is for a simple form of exception handling at the bottom of a function.

systemv/lpstat.c
```diff
@@ -659,7 +659,7 @@ match_list(const char *list,		/* I - List of names */
 
     while (*list && !isspace(*list & 255) && *list != ',')
       list ++;
-  }
+  } while (*list);
```
michaelrsweet, line 662: Coding style - the while should appear on a line after the closing brace...

michaelrsweet (changes requested):
In general this looks good. You need to leave the NULL function checks in place since on macOS looking for the address of a function is a way to detect the presence of an optional ("weak linked") library.

AtariDreams:
All issues addressed!

PR #5955, opened 2021-09-04
Bring configuration improvements from 2.4

AtariDreams:
No actual code changes or new features are made or added. Instead, the scripts have been modernized only with the configuration bug fixes ported over.

The generated scripts are functionally the same, just cleaned up and refactored.


michaelrsweet:
Understand, we *do not* make major changes to the configure script or build system in a patch release. These changes are fine for CUPS 2.4, but *not* for CUPS 2.3.x.

PR #5958, opened 2021-09-07
Clang-tidy fixes

AtariDreams:


cups/testsnmp.c
```diff
@@ -97,13 +97,10 @@ main(int  argc,				/* I - Number of command-line args */
...
 
-  if (!oid)
-  {
-    if (!show_oid(fd, community,  &(host->addr),
-                  walk ? ".1.3.6.1.2.1.43" :
-		         ".1.3.6.1.2.1.43.10.2.1.4.1.1", walk))
+  if (!oid && !show_oid(fd, community, &(host->addr),
+          walk ? ".1.3.6.1.2.1.43" :
+                  ".1.3.6.1.2.1.43.10.2.1.4.1.1", walk))
       return (1);
```
michaelrsweet, line 103: OK, but please fix the indentation.

doc/help/license.html
```diff
@@ -8,7 +8,7 @@
 
 <h1>Apache License</h1>
 
-<p>Version 2.0, January 2004<br></br>
+<p>Version 2.0, January 2004<br/>
```
michaelrsweet, line 11: Not XHTML, so `<br>` is the correct version.

filter/common.c
```diff
@@ -348,26 +348,23 @@ WriteLabelProlog(const char *label,	/* I - Page label */
...
   }
 
-  if (label)
-  {
-    if (classification[0])
-      printf(" - ");
+  if (classification[0])
+    printf(" - ");
 
    /*
```
michaelrsweet, line 354: Please fix the indentation.

filter/common.c
```diff
@@ -348,26 +348,23 @@ WriteLabelProlog(const char *label,	/* I - Page label */
...
 
-    for (ptr = label; *ptr; ptr ++)
-      if (*ptr < 32 || *ptr > 126)
-        printf("\\%03o", *ptr);
-      else
-      {
-	if (*ptr == '(' || *ptr == ')' || *ptr == '\\')
-	  putchar('\\');
+  for (ptr = label; *ptr; ptr++)
+    if (*ptr < 32 || *ptr > 126)
```
michaelrsweet, line 359: For consistency with the current coding guidelines, we now would put braces around each of these blocks. So braces for the `for` loop and for this `if` statement (don't need them for the nested `if` below because there are no compound statements).

tools/ippfind.c
```diff
@@ -336,18 +336,14 @@ main(int  argc,				/* I - Number of command-line args */
...
-            else
-              i ++;
-
-          if (i >= argc)
-          {
-            _cupsLangPrintf(stderr, _("ippfind: Expected semi-colon after %s."),
-                            "--exec");
-            show_usage();
-          }
+            {
```
michaelrsweet, line 341: This logic is different than the original code. We look for a semicolon argument, and show the error if we don't find one.

tools/ippfind.c
```diff
@@ -336,18 +336,14 @@ main(int  argc,				/* I - Number of command-line args */
                                argv + i)) == NULL)
             return (IPPFIND_EXIT_MEMORY);
 
-          while (i < argc)
+          do
```
michaelrsweet, line 339: Braces around the `do` statement please...

michaelrsweet (changes requested):
See comments.

tools/ippfind.c
AtariDreams: Fixed!

AtariDreams:
All issues fixed!

AtariDreams:
All changes finished!

cups/tls-gnutls.c
```diff
@@ -822,10 +822,10 @@ httpLoadCredentials(
 
     httpFreeCredentials(*credentials);
     *credentials = NULL;
+    return -1;
```
michaelrsweet, line 825: Won't this potentially leak "data"?

locale/po2strings.c
```diff
@@ -191,13 +191,11 @@ main(int  argc,				/* I - Number of command-line args */
             cupsFilePrintf(strings, "\"%s\" = \"%s\";\n", msgid, normalize_string((use_msgid || !*msgstr) ? msgid : msgstr, normalized, sizeof(normalized)));
 	}
 
-	if (msgid)
-	  free(msgid);
+  free(msgid);
```
michaelrsweet, line 194: Indentation?

locale/po2strings.c
```diff
@@ -251,14 +248,13 @@ main(int  argc,				/* I - Number of command-line args */
...
-        if (msgstr)
-          free(msgstr);
+       free(msgstr);
 
-	if ((msgstr = strdup(ptr)) == NULL)
-	{
-	  free(msgid);
-	  perror("Unable to allocate msgstr");
-	  return (1);
+       if ((msgstr = strdup(ptr)) == NULL)
```
michaelrsweet, line 253: Indentation?

scheduler/cups-driverd.cxx
```diff
@@ -2400,6 +2400,12 @@ load_ppds(const char *d,		/* I - Actual directory */
   */
 
   dinfoptr = (struct stat *)malloc(sizeof(struct stat));
+  if (!dinfoptr)
+  {
+    fputs("ERROR: [cups-driverd] Unable to allocate memory for directory info!\n",
```
michaelrsweet, line 2405: Don't need the exclamation mark - a period is fine here. Also, if we want to be consistent with some of the existing code you can do the assignment and test in the "if", e.g.:

```
if ((dinfoptr = ...) == NULL)
{
  fputs(...);
  exit(1);
}
```

scheduler/cups-driverd.cxx
```diff
@@ -2621,10 +2627,9 @@ load_ppds_dat(char   *filename,		/* I - Filename buffer */
       {
 	if ((ppd = (ppd_info_t *)calloc(1, sizeof(ppd_info_t))) == NULL)
 	{
-	  if (verbose)
-	    fputs("ERROR: [cups-driverd] Unable to allocate memory for PPD!\n",
-		  stderr);
-	  exit(1);
+    fputs("ERROR: [cups-driverd] Unable to allocate memory for PPD!\n",
```
michaelrsweet, line 2630: Check indentation, period instead of exclamation point.

tools/ippfind.c
```diff
@@ -336,18 +336,15 @@ main(int  argc,				/* I - Number of command-line args */
...
-          if (i >= argc)
+          do
           {
-            _cupsLangPrintf(stderr, _("ippfind: Expected semi-colon after %s."),
-                            "--exec");
-            show_usage();
-          }
+            if (strcmp(argv[i], ";"))
+            {
+              _cupsLangPrintf(stderr, _("ippfind: Expected semi-colon after %s."),
```
michaelrsweet, line 343: Again, we are looking for ";" in argv. You need to loop through the argument list and stop when you see it. If you don't see it, you show an error.

michaelrsweet (changes requested):
Thanks, lots of good stuff here!

See my comments. Mostly code style things but also a few places where the changes alter the logic or might cause a new issue.

backend/snmp.c
```diff
@@ -1050,21 +1045,19 @@ read_snmp_response(int fd)		/* I - SNMP socket file descriptor */
...
-	 /*
+          /*
 	  * Convert the ID to a make and model string...
 	  */
 
-	  backendGetMakeModel((char *)packet.object_value.string.bytes,
-	                      make_model, sizeof(make_model));
-	  if (device->make_and_model)
-	    free(device->make_and_model);
+          backendGetMakeModel((char *)packet.object_value.string.bytes,
```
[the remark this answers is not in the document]
AtariDreams: Had to fix that when I formatted the changed line

cups/http-support.c
```diff
@@ -1355,7 +1355,7 @@ _httpSetDigestAuthString(
     * Follow RFC 2617/7616...
     */
 
-    int		i;			/* Looping var */
+    unsigned		i;			/* Looping var */
```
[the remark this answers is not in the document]
AtariDreams: Not really... there is a reason why array sizes are NSUintegers

cups/tls-darwin.c
```diff
@@ -1808,7 +1808,7 @@ _httpTLSWrite(http_t     *http,		/* I - HTTP connection */
 	       const char *buf,		/* I - Buffer holding data */
 	       int        len)		/* I - Length of buffer */
 {
-  ssize_t	result;			/* Return value */
+  int	result;			      /* Return value */
```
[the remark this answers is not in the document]
AtariDreams: gnutls_record_send does, not the function this particular function calls does though

cups/tls-gnutls.c
AtariDreams: it behaves the exact same as before

tools/ippfind.c
AtariDreams: That's what this code does already?

AtariDreams:
All issues addressed @michaelrsweet

cgi-bin/help-index.c
```diff
@@ -686,17 +686,13 @@ help_delete_node(help_node_t *n)	/* I - Node */
   if (!n)
     return;
 
-  if (n->filename)
-    free(n->filename);
+  free(n->filename);
```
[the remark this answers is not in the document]
AtariDreams: Did!

cgi-bin/var.c
```diff
@@ -396,15 +397,15 @@ cgiSetArray(const char *name,		/* I - Name of variable */
       var->values  = temp;
     }
 
-    if (element >= var->nvalues)
+    if (element < var->nvalues)
+      free((char *)var->values[element]);
```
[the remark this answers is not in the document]
AtariDreams: Did!

AtariDreams:
That's all the changes I wanted to make. Thank you so much!

michaelrsweet (changes requested):
Can't take the changes that remove "& 255" for sure - those were in a separate PR and I've explained why they can't go away.

AtariDreams:
@michaelrsweet ONLY the ones that call the c standard functions for chars, as they were not meant to have the input done with & 255. Everything else I left

michaelrsweet:
@AtariDreams The definition of the ctype macros allows for the value -1 (`EOF` constant) and character values from 0 to 255 (the range of a typical unsigned char with 8-bit bytes).

The `char` data type is (explicitly per the C standards going back to K&R) undefined WRT signedness. The default on some platforms is signed char, and others unsigned char. If char is unsigned, then the "& 255" is not needed. But if char is signed, these ctype macros blow up spectacularly when confronted with UTF-8 text.

Our options are:

1. Keep the "& 255"
2. Write new versions of the macros that explicitly ignore values outside the range of 0 to 127 (although that will have some side-effects for some 8-bit charsets that the ctype macros might otherwise support)
3. Write new versions of the macros that are UTF-8 aware and take a pointer instead of an integer

Given that Apple CUPS is strictly in maintenance mode, I choose option 1.

PR #5962, opened 2021-09-15
Clang tidy 2.0

AtariDreams:
This time, the focus of this fix is on abuse of & 255, especially used in place of casting to unsigned char.

Not to mention, the clib functions for characters already handle the sign extension of chars passed to it.

Finally, I added function header ifdefs.


michaelrsweet:
See my comments on the OpenPrinting CUPS PR you filed. Not getting applied on Apple CUPS...

PR #5972, opened 2021-09-27
fix _cupsGet1284Values

AtariDreams:
Synced from upstream

### DIFF
--- a/cups/options.c
+++ b/cups/options.c
@@ -495,36 +495,30 @@ cupsRemoveOption(
   * Range check input...
   */
 
-  if (!name || num_options < 1 || !options)
-  {
-    DEBUG_printf(("3cupsRemoveOption: Returning %d", num_options));
-    return (num_options);
-  }
-
- /*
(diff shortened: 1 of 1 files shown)


cups/options.c
```diff
@@ -495,36 +495,30 @@ cupsRemoveOption(
...
- /*
-  * Loop for the option...
-  */
-
-  for (i = num_options, option = *options; i > 0; i --, option ++)
-    if (!_cups_strcasecmp(name, option->name))
-      break;
-
-  if (i)
+  if (name && num_options > 0 && options)
```
michaelrsweet, line 498: This is not part of the original 1284 fix...

michaelrsweet (changes requested):
You've included a bunch of additional changes of your own, not just the original fix for `_cupsGet1284Values`.

michaelrsweet:
Pushed the changes (and ONLY the changes) from OpenPrinting CUPS myself.

PR #5985, opened 2021-10-14
Avoid a crash when an Avahi or mDNS printer has neither device-uri nor printer-uri-supported set.

AlaskanEmily:
This was found in https://bugzilla.mozilla.org/1691347


AtariDreams:
@michaelrsweet

Issue #6076, opened 2022-10-31
Can access GUI admin interface without authentication, but Find New Printers requires an authenticated user. 


Dominic-Mayers:
I am using cups GUI admin interface on a fresh installation of Ubuntu 22.10. If I visit `127.0.0.1:631/admin` I do not have to authenticate, but without authentication, if I click on `Find New Printers`, no printers are found. Yet I see a printer with `avahi-browse -t _ipp._tcp` and also directly in `Settings->Printers`. If I artificially force a login as root  by clicking on `Add Printer`, but don't proceed ahead and instead go back to `Administration` and click on `Find new Printers`, then the printer is found. BTW, it might not be related, but if I use `localhost:631` instead of `127.0.0.1:631`, the login as root fails and the login window  comes back. Back to the situation with `127.0.0.1:631`, if I modify the configuration file `cupsd.conf` and replace 
```
# Restrict access to the admin pages...
<Location /admin>
  Order allow,deny
</Location>
``` 
with
```
# Restrict access to the admin pages...
<Location /admin>
  AuthType Default
  Require user @SYSTEM
  Order allow,deny
</Location>
``` 
then I cannot reach the admin page without first seeing the login window that allows me to login as root and the problem is solved. There might be a problem if the root user does not have a password, but this problem exists for `Add Printer` already.

Dominic-Mayers:
This picture illustrates that cups asks the user for a root password : ![cups_asking_for_root_password](https://user-images.githubusercontent.com/5099917/199131959-17f9a291-ed7f-4be6-9ecf-ab4e6196965f.png)  I believe that this is in itself a bigger issue. It's related, because why should we care which user is using the interface, given that all users that have access to the computer have access to the underlying commands.

PR #6077, opened 2022-11-02
Change cupsd.conf to force authentication when accessing administration

Dominic-Mayers:
With the current default configuration, someone can request to Find New Printers without being authenticated and with no way to be authenticated. Without authentication cups does not detects printers on the network. The person might not realize that it's because authentication is missing. It seems natural to always offer to authenticate when accessing the administration page. See https://github.com/apple/cups/issues/6076 .


AtariDreams:
@michaelrsweet

PR #6084, opened 2022-11-11
Fix backchannel status checking bugs

AtariDreams:
Checking for merely (status < 0) is a bug because the check for timeout is deliberate.

Note that you have an if (timeout < 0) at the beginning

So if the timeout is set to a value, it should check for that value.

The condition on the while is also suspicious, though.

It should be while (status < 0 && (errno == EINTR || errno == EAGAIN))

Co-Authored-By: aaaaaa123456789 <aaaaaa123456789@acidch.at>


AtariDreams:
@salgernon @michaelrsweet Please take a look!

PR #6087, opened 2022-11-16
[Mistake Bug-Fix]: Fix cupsfilter by setting both array elements to -1

AtariDreams:
Right now, we set filterfds[1 - current][0] = -1; twice, when we should set filterfds[1 - current][1] to -1 as well


AtariDreams:
@michaelrsweet Please could you accept this quickie

AtariDreams:
@salgernon

PR #6147, opened 2023-09-17
Add jobs' original hostname indication in the Jobs table

StarterCraft:
Not all drivers allow to view the job's origin hostname, they often only provide the username. This is for fixing the issue.


StarterCraft:
I am using a Canon MF4550d MFU. Upon installation on Windows, there are two different drivers I can use: Microsoft IPP Class Driver and Canon MF4550D URFII LT Driver.

The first one, should it be called Microsoft driver for short, reports usernames in `HOSTNAME\USERNAME` format. But as the standard driver, it only has basic functionality and so I prefer to use the Canon driver.

The other driver (Canon MF4550d URFII LT) reports just the user name. It can be a problem if an user uses the same username on different computers, which often happens on Windows PCs.

This is the reasoning behind the changes I made. It would be great to also add an option to choose whether the hostname should be displayed or not. But my knowledge in C++ doesn't allow me to do that.

